PR #7, opened 2023-06-20
CMM-817: Added raisefault for 400 error

ChewieCB:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/proxies/live/apiproxy/policies/RaiseFault.400BadRequest.xml
+++ b/proxies/live/apiproxy/policies/RaiseFault.400BadRequest.xml
@@ -1,9 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <RaiseFault async="false" continueOnError="false" enabled="true" name="RaiseFault.400BadRequest">
+    <DisplayName>RaiseFault.400BadRequest</DisplayName>
+    <Properties/>
     <FaultResponse>
         <Set>
-            <Payload contentType="text/plain"/>
+            <Headers/>
+            <Payload variablePrefix="%" variableSuffix="#">
+            {
+                "errors" : %errors#
+            }
+            </Payload>
             <StatusCode>400</StatusCode>
-            <ReasonPhrase>Bad Request</ReasonPhrase>
+            <ReasonPhrase>Bad request</ReasonPhrase>
         </Set>
     </FaultResponse>
     <IgnoreUnresolvedVariables>true</IgnoreUnresolvedVariables>
--- a/proxies/live/apiproxy/proxies/default.xml
+++ b/proxies/live/apiproxy/proxies/default.xml
@@ -69,6 +69,15 @@
   <RouteRule name="communications-manager-target">
     <TargetEndpoint>communications-manager-target</TargetEndpoint>
   </RouteRule>
+
+  <FaultRules>
+   <FaultRule name="missing_parameter">
+      <Step>
+        <Name>RaiseFault.400BadRequest</Name>
+      </Step>
+      <Condition>(fault.name = "FailedToResolveAPIKey")</Condition>
+  </FaultRule>
+  </FaultRules>
   <DefaultFaultRule>
     <Step>
       <Name>AssignMessage.Errors.CatchAllMessage</Name>
--- a/proxies/live/apiproxy/policies/JavaScript.MessageBatches.Create.Validate.xml
+++ b/proxies/live/apiproxy/policies/JavaScript.MessageBatches.Create.Validate.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<Javascript async="false" continueOnError="false" enabled="true" timeLimit="200" name="JavaScript.MessageBatches.Create.Validate">
+    <DisplayName>JavaScript.MessageBatches.Create.Validate</DisplayName>
+    <Properties/>
+    <ResourceURL>jsc://MessageBatches.Create.Validate.js</ResourceURL>
+</Javascript>
--- a/proxies/live/apiproxy/resources/jsc/MessageBatches.Create.Validate.js
+++ b/proxies/live/apiproxy/resources/jsc/MessageBatches.Create.Validate.js
@@ -1,0 +1,220 @@
+// Regex to validate the id is a valid UUID
+const uuidRegex = /^[0-9a-f]{8}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{4}\b-[0-9a-f]{12}$/i
+const nhsNumberRegex = /^[0-9]{10}$/i
+const dobRegex = /^[0-9]{4}-[0-9]{2}-[0-9]{2}$/i
+
+var content = context.getVariable("request.content")
+var errors = []
+
+var all
+try {
+  all = JSON.parse(content)
+} catch(e) {
+  pushError(invalidError("/"))
+}
+
+if (all) {
+  var seenMessages = []
+  var data = all.data;
+
+  var pointer
+
+  // $.data
+  pointer = "/data"
+  if (typeof data === "undefined") {
+    pushError(missingError(pointer))
+  } else if (data === null) {
+    pushError(nullError(pointer))
+  } else if (typeof data !== "object") {
+    pushError(invalidError(pointer))
+  } else {
+
+    // $.data.type
+    pointer = "/data/type"
+    if (typeof data.type === "undefined") {
+      pushError(missingError(pointer))
+    } else if (data.type === null) {
+      pushError(nullError(pointer))
+    } else if (data.type !== "MessageBatch") {
+      pushError(invalidError(pointer))
+    }
+
+    // $.data.attributes
+    pointer = "/data/attributes"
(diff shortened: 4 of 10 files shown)


proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+//   "attributes/routingPlanId",
+//   "attributes/batchReference",
+//   "attributes/recipients",
+//   "attributes/recipients/0/recipientReference",
+//   "attributes/recipients/0/nhsNumber",
+//   "attributes/recipients/0/dateOfBirth",
+//   "attributes/recipients/0/personalisation",
+// ]
+// var requestParameters = context.getVariable("")
+var seenRecipients = []
```
PhilipSkinner, line 18: I thought it may be faster to do this lookup in a dict, but it isn't! The time to check if a value is within an array or within a dict is the same on node 16.

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+var errors = []
```
PhilipSkinner, line 19: Its possible that we may output a response body that is actually larger than the request that came in, if there is an error with every entry in the submission.

I believe we should limit this to return the first 100 errors.

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+if (data) {
+  if (data === null || data === undefined) {
+    title = "Property cannot be null"
+    detail = "The property at the specified location is required, but a null value was passed in the request."
+    errors.push(createErrorObject(title, detail, "/data"))
+  } else {
+
+    // $.data.type
+    pointer = "/data/type"
+    if (data.type) {
```
PhilipSkinner, line 35: For existence of a property use `typeof data.type === 'undefined'` - this is our check for if the property is undefined, `data.type === null` for our null check and finally we can check for value validity.

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+var errors = []
+
+var title
+var detail
+var pointer
+
+// $.data
+if (data) {
+  if (data === null || data === undefined) {
+    title = "Property cannot be null"
```
PhilipSkinner, line 28: can probably move these out to a set of constant values then reference them instead

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+          errors.push(createErrorObject(title, detail, pointer))
+        } else {
+          if (data.attributes.recipients.length() < 1) {
+            title = "Too few items"
+            detail = "The property at the specified location contains too few items."
+            errors.push(createErrorObject(title, detail, pointer))
+          } else {
+
+            // $.data.attributes.recipients.x
+            for (var i = 0; i < data.attributes.recipients.length(); i++) {
```
PhilipSkinner, line 99: `data.attributes.recipients.forEach((recipient, index) => {
    ...
});`

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+              pointer = "/data/attributes/recipients/" + i
+              var item = data.attributes.recipients[i]
+              if (!item || item === null || item === undefined) {
+                title = "Missing property"
+                detail = "The property at the specified location is required, but was not present in the request."
+                errors.push(createErrorObject(title, detail, pointer))
+              } else {
+
+                // $.data.attributes.recipients.x.recipientReference
+                pointer = "/data/attributes/recipients/" + i + "/recipientReference"
```
PhilipSkinner, line 109: Interpolation instead:

`/data/attributes/recipients/${i}/recipientReference`

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+var seenRecipients = []
+var errors = []
+
+var title
+var detail
+var pointer
+
+// $.data
+if (data) {
+  if (data === null || data === undefined) {
```
PhilipSkinner, line 27: typeof data === 'undefined'

plus can return early here to reduce nesting

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
...
+// var requestParameters = context.getVariable("")
+var seenRecipients = []
+var errors = []
+
+var title
+var detail
+var pointer
+
+// $.data
+if (data) {
```
PhilipSkinner, line 26: This means the if statement on the line below never gets executed, can probably remove this outer statement and rely on the check on line 27.

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
ChewieCB: Yeah unfortunately the javascript engine in Apigee doesn't support some features, so we're stuck with basic string concatenation

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,210 @@
+// Regex to validate the id is a valid UUID
+const uuidRegex = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi
+const nhsNumberRegex = /^[0-9]{10}$/gi
+const dobRegex = /^[0-9]{4}-[0-9]{2}-[0-9]{2}$/gi
+
+var data = context.getVariable("request.content.data")
```
ChewieCB, line 6: `context.getVariable("")` and `context.setVariable("")` is the main method of changing flow variables within the apigee flow from within a JS policy.

proxies/live/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,200 @@
...
+//     property = property.split('/');
+//     for (var i = 0; i < property.length; i++) {
+//         if(typeof obj[property[i]] == 'undefined')
+//             return defval;
+//         obj = obj[property[i]];
+//     }
+//     return obj;
+// }
+
+function createErrorObject(title, detail, pointer) {
```
ChewieCB, line 143: I'll probably move all of these error message generating functions into an external library file for the sake of conciseness

proxies/live/apiproxy/policies/RaiseFault.400BadRequest.xml
```diff
@@ -1,9 +1,15 @@
...
+    <Properties/>
     <FaultResponse>
         <Set>
-            <Payload contentType="text/plain"/>
+            <Headers/>
+            <Payload variablePrefix="%" variableSuffix="#">
+              %errors#
+            </Payload>
             <StatusCode>400</StatusCode>
-            <ReasonPhrase>Bad Request</ReasonPhrase>
```
PhilipSkinner, line 6: Should be Bad Request rather than Missing property.

proxies/live/apiproxy/proxies/default.xml
```diff
@@ -69,6 +69,15 @@
   <RouteRule name="communications-manager-target">
     <TargetEndpoint>communications-manager-target</TargetEndpoint>
   </RouteRule>
+
+  <FaultRules>
+   <FaultRule name="missing_parameter">
+      <Step>
+        <Name>RaiseFault.400BadRequest</Name>
```
PhilipSkinner, line 76: Probably needs a different error, we can leave it as it is now as our focus is on sandbox.

proxies/sandbox/apiproxy/resources/jsc/SetErrorVars.js
```diff
@@ -0,0 +1,202 @@
...
+          errors.push(invalidError(pointer))
+        } else if (message.recipient === null) {
+          errors.push(nullError(pointer))
+        } else {
+
+          // $.data.attributes.messages.x.recipient.nhsNumber
+          pointer = "/data/attributes/messages/" + index + "/recipient/nhsNumber"
+          if (typeof message.recipient.nhsNumber === 'undefined') {
+            errors.push(missingError(pointer))
+          } else if (message.recipient.nhsNumber === null || message.recipient.nhsNumber === "") {
```
PhilipSkinner, line 107: I believe the logic here is invalid, the null error should only throw when its null - an empty string is a none null value and should instead throw the invalid error. This should be the same across all fields.
* undefined == missing error
* null = null error
* invalid value (or type) == invalid error

PhilipSkinner (approved):
Looks good now, accepting.

PR #48, opened 2023-07-03
Update communications-manager.yaml

LEUNHU:
Added content for 'Technology' subheader.

## Summary
* Routine Change

## Reviews Required
* [x] Dev

## Review Checklist

### DIFF
--- a/specification/communications-manager.yaml
+++ b/specification/communications-manager.yaml
@@ -19,7 +19,8 @@ info:
     The Communications Manager API is currently not in an integration or production environment.
 
     ## Technology 
-    TBC
+    This API is a [REST-based](https://digital.nhs.uk/developer/guides-and-documentation/our-api-technologies#basic-rest) API.
+    We follow the [JSON:API](https://jsonapi.org/) standard for our request and response schemas. 
 
     ## Network access
     TBC


specification/communications-manager.yaml
```diff
@@ -19,7 +19,8 @@ info:
     The Communications Manager API is currently not in an integration or production environment.
 
     ## Technology 
-    TBC
+    This API is a [REST-based](https://digital.nhs.uk/developer/guides-and-documentation/our-api-technologies#basic-rest) API.
```
PhilipSkinner, line 22: Enter an extra blank line in between these values so we get them broken up when this renders, see:

image

PhilipSkinner (changes requested):

specification/communications-manager.yaml
LEUNHU: @PhilipSkinner I've (hopefully!) added another line using > rather than \n as discussed.

LEUNHU: Amended as requested now - thanks for the info.

PhilipSkinner: Thanks, this looks like this now:

image

PR #58, opened 2023-07-03
CCM-954

PhilipSkinner:
## Summary
Adds documentation onto the POST /v1/message-batches endpoint.

Adds documentation should a 201 response be generated.

## Reviews Required
* [ ] Dev
* [ ] Test
* [ ] Tech Author
* [x] Product Owner

## Review Checklist

### DIFF
--- a/specification/communications-manager.yaml
+++ b/specification/communications-manager.yaml
@@ -49,6 +49,25 @@ paths:
 
         Use this endpoint to send a batch of messages to 1 or more NHS patients.
 
+        ### References
+
+        You must provide two reference values within the payload to this endpoint:
+
+        * A message batch reference
+        * A per message reference
+
+        Both of these references must be a [Universal Unique Identifier (UUID)](https://en.wikipedia.org/wiki/Universally_unique_identifier).
+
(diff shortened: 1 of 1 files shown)


specification/communications-manager.yaml
```diff
@@ -49,6 +49,25 @@ paths:
...
         Use this endpoint to send a batch of messages to 1 or more NHS patients.
 
+        ### References
+
+        You must provide two reference values within the payload to this endpoint:
+
+        * A message batch reference
+        * A per message reference
+
+        Both of these references must be UUIDs.
```
LEUNHU, line 59: Given UUID is an acronym, should we use the full term followed by (UUIDs)

LEUNHU (changes requested):
1 comment on use of acronyms

specification/communications-manager.yaml
PhilipSkinner: Good shout, I'll make this part better.

PhilipSkinner: I've given the full term, the acronym plus I've linked to wikipedia so the consumer can find more information on UUIDs if they have never used them before.

LEUNHU (approved):

PR #60, opened 2023-07-04
CCM-964

PhilipSkinner:
## Summary
Corrects example routing plan id to equal one of the valid
routing plans in sandbox.

Adds new `Links_Error` schema to dedupe the links section within
each error response.

Adds a type of string onto all enums.

Enforces that all scalar types must have an example value set.

Adds example values for scalars that are missing them.

## Reviews Required
* [ ] Dev
* [ ] Test
* [ ] Tech Author
* [x] Product Owner

## Review Checklist

### DIFF
--- a/redocly.yaml
+++ b/redocly.yaml
@@ -1,2 +1,3 @@
 rules:
   info-license: off
+  scalar-property-missing-example: error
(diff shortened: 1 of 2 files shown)


PhilipSkinner:
Before the change errors with enums would look like this:

image

PhilipSkinner:
After the change the values are rendering correctly:

image

The improvements to the spec and the checking will ensure that we don't miss example values on scalars from this point onwards.

LEUNHU (approved):

PR #80, opened 2023-07-06
CMM-983: Fixed message recipient validation

ChewieCB:
The recipient attribute of each message was not dependent on successful validation of the parent message attribute - hence we were getting extra recipient errors when the message was invalid.


PhilipSkinner:
Closing, duplicated PR.

PR #141, opened 2023-07-21
CCM-967: Added proxy check before test runs

ChewieCB:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/azure/templates/run-tests.yml
+++ b/azure/templates/run-tests.yml
@@ -36,6 +36,21 @@ steps:
     displayName: Setup pytests
     condition: always()
 
+  - bash: |
+      proxyResponse=$(
+      curl --request GET 'https://sandbox.api.service.nhs.uk/comms/v1/message-batches' \
+           --header 'Content-Type: application/vnd.api+json'
+      )
+      echo "$proxyResponse"
+      errorResponse=$($proxyResponse | jq ".errors[0]")
+      if [[ $($errorResponse | jq ".status") == "404" && $($errorResponse | jq ".code") == "CM_NOT_FOUND" ]]; then
+        exit 0
+      fi
+      exit 1
+    displayName: Wait for proxy to deploy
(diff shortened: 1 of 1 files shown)


azure/templates/run-tests.yml
```diff
@@ -36,6 +36,27 @@ steps:
...
+      code=$(echo $errorResponse | jq -r ".code")
+
+      if [[ ($status == "404" && $code == "CM_NOT_FOUND") || ($status == "401" && $code == "CM_DENIED") ]]; then
+          exit 0
+      else
+          exit 1
+      fi
+    workingDirectory: $(Pipeline.Workspace)/s/$(SERVICE_NAME)/$(SERVICE_ARTIFACT_NAME)
+    displayName: Check proxy is deployed
+    retryCountOnTaskFailure: 3
```
PhilipSkinner, line 58: Will this wait between retries, if so how long does it wait?

I think this value may need to be higher to cater for when its taking some time for the proxy to have been stood up.

PhilipSkinner (changes requested):

azure/templates/run-tests.yml
ChewieCB: From the azure docs it looks like it has exponential backoff built-in, starting with a few seconds and increasing per retry. They don't give a specific timing value though, I'd expect it varies due to performance.

> The failing task retries in seconds. The wait time between each retry increases after each failed attempt.

PhilipSkinner: Excellent, that sounds ideal - I mean not about the azure docs not being clear but the functionality sounds perfect.

PhilipSkinner (approved):
Approved!

PR #153, opened 2023-07-26
CCM-1228 - validation performance

PhilipSkinner:
## Summary

Adds script that allows for performance testing of create message validation scenarios.

Increases the performance on the validation script.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/proxies/shared/resources/jsc/MessageBatches.Create.Validate.js
+++ b/proxies/shared/resources/jsc/MessageBatches.Create.Validate.js
@@ -18,13 +18,13 @@ const isUndefined = (val) => {
     return typeof val === "undefined";
 }
 
-if (all) {
-  var seenMessages = {};
-  var data = all.data;
+const validate = () => {
+  if (all) {
+    var seenMessages = {};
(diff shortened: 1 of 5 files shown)


proxies/shared/resources/jsc/MessageBatches.Create.Validate.js
```diff
@@ -14,154 +14,162 @@ try {
...
 }
 
-
 function pushError(error) {
   // Limit the amount of errors returned to 100 entries
   if (errors.length <= 100) {
-    errors.push(error)
+    errors.push(error);
+  } else {
+    throw new Error("Maximum number of error entries reached.");
```
PhilipSkinner, line 169: Exceptions can have quite a big overhead- we can probably add a check for each message instead which just terminates the for each loop instead.

If you can show that the exception doesn't add a massive overhead then I'll accept this!

markpullan (approved):

PR #196, opened 2023-08-18
CCM-1405: Allow requests with no Accept Header

seanmarjason-nhs:
## Summary
* Routine Change
Allow requests when Accept Header is not defined.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/specification/communications-manager.yaml
+++ b/specification/communications-manager.yaml
@@ -49,6 +49,14 @@ info:
 
     You can control which `Content-Type` is returned by using the `Accept` header.
 
+    The `Accept` header can contain the following values:
+
+    * `*/*`
+    * `application/json`
+    * `application/vnd.api+json`
+
+    Where no `Accept` header is present, this will default to `application/vnd.api+json`
+
     ### Request Content Types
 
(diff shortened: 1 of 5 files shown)


specification/communications-manager.yaml
```diff
@@ -377,6 +377,8 @@ paths:
             * `*/*`
             * `application/json`
             * `application/vnd.api+json`
+
+            Where no `Accept` header is present, this will default to `application/vnd.api+json`
```
PhilipSkinner, line 381: I think we should expand on the documentation further up in the API description where we mention the acceptable types the server can generate, maybe adding in:

* A list of valid `Accept` header values
* Information on which will be default when no `Accept` header is present

PhilipSkinner (changes requested):
Small documentation change please

specification/communications-manager.yaml
PhilipSkinner: Thanks @seanmarjason-nhs !

PhilipSkinner (approved):
Approved

PR #260, opened 2023-10-02
CCM-1616 - Service bans

PhilipSkinner:
## Summary

Adds a new policy that allows for service ban responses from the backend solution to trigger a custom 403 response within the API proxy.

Ensures that the sandbox mock application can throw errors in the same way as the backend application, with test coverage over it and the policy being run.

Updates the documentation for the proxy to include the new logic around the handling of 403 & 403 service bans.

Adds the 403 service ban information to the API specification.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/sandbox/handlers.js
+++ b/sandbox/handlers.js
@@ -93,6 +93,12 @@ const duplicateTemplates = [
 ];
 
 async function batch_send(req, res, next) {
+    if (req.headers["authorization"] === "banned") {
+        sendError(res, 403, "Request rejected because client service ban is in effect");
+        next();
+        return;
+    }
+
     if (!req.body) {
         sendError(res, 400, "Missing request body");
         next();
--- a/tests/sandbox/create_message_batches/test_service_ban.py
+++ b/tests/sandbox/create_message_batches/test_service_ban.py
@@ -1,0 +1,84 @@
+import requests
+import pytest
+from lib import Assertions, Generators
+from lib.constants import CORRELATION_IDS
+
+FORBIDDEN_TOKEN = {
+    "Authorization": "Bearer ClientNotRecognised"
+}
+
+
+@pytest.mark.sandboxtest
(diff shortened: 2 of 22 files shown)


tests/sandbox/create_message_batches/test_service_ban.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.mark.sandboxtest
+@pytest.mark.parametrize("correlation_id", CORRELATION_IDS)
+def test_prefer_403_service_ban(nhsd_apim_proxy_url, correlation_id):
+    """
+    .. py:function:: Scenario: An API consumer wants to test the service ban \
+        error message on the sandbox environment
+
+        | **Given** the API consumer wants to test the service ban response
+        | **When** a request is submitted with a ``Prefer`` header of ``code=403.11``
```
Kennedy1265, line 57: I'm partial to the use of Given, When Then as:

Given - Set up - what is different in the setup of the test 
When - Action - what is the user doing
Then - Outcome - what is the result of the action with the differences provided in the set up

So maybe something more like:
```
**Given** the API consumer provides a code 403.11 prefer header
**When** the request is submitted
**Then** the response returns a 403 service ban error
```

PhilipSkinner: sorry, I got that wrong, I'll get it updated

Kennedy1265 (approved):

sandbox/handlers.js
```diff
@@ -93,6 +93,12 @@ const duplicateTemplates = [
 ];
 
 async function batch_send(req, res, next) {
+    if (req.headers["authorization"] === "banned") {
```
Ian-Hodges, line 96: Do we need this now we have the **prefer: code=403.1** logic in apigee?

PR #268, opened 2023-10-04
CCM-1866: Handle unexpected 504

Kennedy1265:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

* [x] I have reviewed the changes in this PR and they fill all or part of the acceptance criteria of the ticket, and the code is in a mergeable state.

### DIFF
--- a/tests/development/content_types/test_content_type_negotiation.py
+++ b/tests/development/content_types/test_content_type_negotiation.py
@@ -1,6 +1,7 @@
 import requests
 import pytest
-from lib.constants import METHODS, UNEXPECTED_429
+from lib.constants import METHODS
+from lib import Error_Handler
 
 DEFAULT_CONTENT_TYPE = "application/vnd.api+json"
 ACCEPT_HEADERS = [
@@ -38,7 +39,6 @@ def test_application_response_type(nhsd_apim_proxy_url, accept_headers, method, 
         **accept_headers.get("headers")
     })
 
-    if resp.status_code == 429:
-        raise UNEXPECTED_429
+    Error_Handler.handle_retry(resp)
 
     assert resp.headers.get("Content-Type") == accept_headers.get("expect")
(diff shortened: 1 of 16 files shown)


tests/development/content_types/test_content_type_negotiation.py
```diff
@@ -41,4 +41,7 @@ def test_application_response_type(nhsd_apim_proxy_url, accept_headers, method,
     if resp.status_code == 429:
```
PhilipSkinner, line 41: I wonder if we should move this out into a helper then we can call it with the `resp` value - that way we only have the retry logic in a single place?

I'm thinking we have 3 functions - 

* `handle_retry(resp)`
* `handle_429_retry(resp)`
* `handle_504_retry(resp)`

; where the `handle_retry` func executes the two other functions in order. The reason for this is that in certain places we want a 429 to be thrown but still want to retry on a 504, and vice versa.

tests/lib/assertions.py
```diff
@@ -1,12 +1,15 @@
-from .constants import CORS_METHODS, CORS_MAX_AGE, CORS_ALLOW_HEADERS, CORS_EXPOSE_HEADERS, CORS_POLICY, UNEXPECTED_429
+from .constants import *
```
PhilipSkinner, line 1: Sonar won't like this, if you want access to all constants do `import lib.constants as constants` then access the constants as `constants.CORS_METHODS` etc

PhilipSkinner:
Missing 504 checks on the sandbox 429 tests

PhilipSkinner (changes requested):
Please see prior comments

tests/test_endpoints.py
```diff
@@ -45,6 +45,9 @@ def test_wait_for_ping(nhsd_apim_proxy_url):
     if resp.status_code == 429:
```
PhilipSkinner, line 45: We should import the Error_Handler here for these tests instead of having the same logic here

tests/test_endpoints.py
```diff
@@ -7,7 +7,8 @@
 import pytest
 import time
 from os import getenv
-from lib.constants import UNEXPECTED_429
+from lib.constants import UNEXPECTED_429, UNEXPECTED_504
```
Ian-Hodges, line 10: Very minor, these constants are no longer used.

Ian-Hodges (approved):
Looks good to me and Phil's comments have been addressed.

PhilipSkinner (approved):

PR #382, opened 2024-01-02
CCM-2152: Remove aws headers from responses

TheWorldWideWebb:
## Summary
This work aims to remove amazon headers from the response, Ive taken a regex approach to pattern match on x-amz. I have extended the testing assertion library to assert on 3 different types on headers which also covers the majority of edge cases e.g. casing and amzn type headers.

## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 1 reviews received

### DIFF
--- a/proxies/shared/resources/jsc/SetResponseDefaults.js
+++ b/proxies/shared/resources/jsc/SetResponseDefaults.js
@@ -21,9 +21,10 @@ context.setVariable("error.header.X-Content-Type-Options", "nosniff");
 // remove aws headers
 const headers = context.getVariable("response.header");
 const headerRegex = /^x-amz-/;
-Object.keys(headers).forEach(function (headerKey) {
-  if (headerRegex.test(headerKey.toLowerCase())) {
-    context.removeVariable("response.header." + headerKey);
+headers.forEach(header => {
+  if (headerRegex.test(header.key.toLowerCase())) {
+    context.removeVariable("response.header." + header.key);
+    context.removeVariable("error.header." + header.key);
   }
 });
 
--- a/proxies/shared/resources/jsc/SetResponseDefaults.js
+++ b/proxies/shared/resources/jsc/SetResponseDefaults.js
@@ -21,12 +21,12 @@ context.setVariable("error.header.X-Content-Type-Options", "nosniff");
 // remove aws headers
 const headers = context.getVariable("response.header");
 const headerRegex = /^x-amz-/;
-headers.forEach(function (header)  {
+for (let header of headers) {
   if (headerRegex.test(header.key.toLowerCase())) {
     context.removeVariable("response.header." + header.key);
     context.removeVariable("error.header." + header.key);
   }
-});
+};
 
 // format errors and response objects
 try {
--- a/proxies/shared/resources/jsc/SetResponseDefaults.js
+++ b/proxies/shared/resources/jsc/SetResponseDefaults.js
@@ -21,7 +21,7 @@ context.setVariable("error.header.X-Content-Type-Options", "nosniff");
 // remove aws headers
 const headers = context.getVariable("response.header");
 const headerRegex = /^x-amz-/;
-headers.forEach(header => {
+headers.forEach(function (header)  {
   if (headerRegex.test(header.key.toLowerCase())) {
     context.removeVariable("response.header." + header.key);
     context.removeVariable("error.header." + header.key);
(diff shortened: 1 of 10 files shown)


tests/sandbox/headers/test_x_amz_removal.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.mark.sandboxtest
+@pytest.mark.parametrize("method", TEST_METHODS)
+@pytest.mark.parametrize("endpoints", VALID_ENDPOINTS)
+def test_request_with_x_amz_is_removed(nhsd_apim_proxy_url, endpoints, method):
+
+    resp = getattr(requests, method)(f"{nhsd_apim_proxy_url}/{endpoints}", headers={
+        "Accept": "*/*",
+        "Origin": "https://my.website"
```
Kennedy1265, line 15: Looks like you've created an ORIGIN constant that could be used here, there's also a couple other instances in the tests that directly set https://my.website that could be replaced as well under cors tests if you're going this route

TheWorldWideWebb: Made consistent across tests

Kennedy1265 (approved):
Testing looks good

proxies/shared/resources/jsc/SetResponseDefaults.js
```diff
@@ -18,6 +18,17 @@ context.setVariable("error.header.Cache-Control", "no-cache, no-store, must-reva
 context.setVariable("response.header.X-Content-Type-Options", "nosniff");
 context.setVariable("error.header.X-Content-Type-Options", "nosniff");
 
+// remove aws headers
+
```
PhilipSkinner, line 22: remove the extra `\n` please

tests/lib/constants/constants.py
```diff
@@ -26,7 +26,9 @@
 
 TOKENS = [None, "Bearer xyzcba", "Bearer", "junk"]
 METHODS = ["get", "post", "put", "patch", "delete", "head", "options"]
+TEST_METHODS = ["get", "post", "put", "delete"]
```
PhilipSkinner, line 29: We should be using the existing set of `METHODS` for these tests I believe.

tests/production/headers/test_cors.py
```diff
@@ -1,8 +1,7 @@
 import requests
 import pytest
 from lib import Assertions, Authentication
-from lib.constants.constants import PROD_URL, VALID_ENDPOINTS
-
+from lib.constants.constants import PROD_URL, VALID_ENDPOINTS, ORIGIN
 
 METHODS = ["GET", "POST", "PUT", "PATCH", "DELETE"]
 TEST_METHODS = ["get", "post", "put", "delete"]
```
PhilipSkinner, line 7: hmmm, why do we have these two lists of methods?

PhilipSkinner (changes requested):
Please see comments, once resolved will give this a tick!

tests/lib/constants/constants.py
TheWorldWideWebb: I would only be inclined to agree if someone could give a reason as to why they are separate currently, or if we can agree as a consensus that we should migrate them to these constants.

tests/production/headers/test_cors.py
TheWorldWideWebb: As above, there existence is unclear

proxies/shared/resources/jsc/SetResponseDefaults.js
```diff
@@ -18,6 +18,17 @@ context.setVariable("error.header.Cache-Control", "no-cache, no-store, must-reva
...
+const headerNames = (context.getVariable("response.headers.names") + "").slice(1, -1).split(', ');
```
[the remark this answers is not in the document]
TheWorldWideWebb: Is that driven from the backend lambdas populating them?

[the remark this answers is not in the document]
TheWorldWideWebb: I did find this which would suggest we dont need to interact with the error headers at all but let me know if this error object is not the same thing

https://docs.apigee.com/api-platform/reference/variables-reference#error
https://docs.apigee.com/api-platform/fundamentals/fault-handling

tests/lib/constants/constants.py
TheWorldWideWebb: Ive tried running the tests with the additional methods but was met with failures for the options method

proxies/shared/resources/jsc/SetResponseDefaults.js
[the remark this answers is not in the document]
PhilipSkinner: Yeah I think you're right, I'm mistaken about this - your code looks good and we can see it working with a backend thrown exception as it correctly removes the AWS headers from it.

PhilipSkinner (approved):
Approved!

PR #395, opened 2024-01-08
CCM-2327: send batch - expose routing plan

beoh1:
## Summary

Return the routing plan id and version on the send batch endpoint.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 1 reviews received

### DIFF
--- a/tests/sandbox/message_batches/create_message_batches/test_success.py
+++ b/tests/sandbox/message_batches/create_message_batches/test_success.py
@@ -9,12 +9,12 @@ VALID_ROUTING_PLAN_ID = [
     "49e43b98-70cb-47a9-a55e-fe70c9a6f77c",
     "b402cd20-b62a-4357-8e02-2952959531c8",
     "936e9d45-15de-4a95-bb36-ae163c33ae53",
-    "9ba00d23-cd6f-4aca-8688-00abc85a7980"
+    "9ba00d23-cd6f-4aca-8688-00abc85a7980",
 ]
 
 
 @pytest.mark.sandboxtest
-@pytest.mark.parametrize('accept_headers', constants.VALID_ACCEPT_HEADERS)
+@pytest.mark.parametrize("accept_headers", constants.VALID_ACCEPT_HEADERS)
 def test_201_message_batch_valid_accept_headers(nhsd_apim_proxy_url, accept_headers):
     """
     .. include:: ../../partials/happy_path/test_201_message_batch_valid_accept_headers.rst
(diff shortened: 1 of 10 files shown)


tests/sandbox/message_batches/create_message_batches/test_success.py
```diff
@@ -69,31 +80,41 @@ def test_201_message_batch_valid_nhs_number(nhsd_apim_proxy_url):
...
-            "Content-Type": "application/json"
-        }, json=data
+    resp = requests.post(
+        f"{nhsd_apim_proxy_url}{MESSAGE_BATCHES_ENDPOINT}",
+        headers={"Accept": "application/json", "Content-Type": "application/json"},
+        json=data,
+    )
+    Assertions.assert_201_response(
+        resp,
+        data["data"]["attributes"]["messageBatchReference"],
```
TheWorldWideWebb, line 115: Non blocking.

Might be worth moving these into variables, its quite a horrible thing to read

tests/sandbox/message_batches/create_message_batches/test_success.py
```diff
@@ -69,31 +80,41 @@ def test_201_message_batch_valid_nhs_number(nhsd_apim_proxy_url):
     .. include:: ../../partials/happy_path/test_201_message_batch_valid_nhs_number.rst
     """
     data = Generators.generate_valid_create_message_batch_body("sandbox")
```
TheWorldWideWebb, line 82: Again, optional. 

Is it worth moving this generation of data into a pre step. The consequence I think is that you'd have to spread it everywhere instead. Let me know your thoughts.

PhilipSkinner: The issue with doing that is that we have idempotency on the API thats related to the batch reference within this message - so it needs to be generated on a per test invocation basis or we'll end up getting back the deduped response.

beoh1: Given the data is different for each test I prefer as is rather than having to spread tbh

tests/sandbox/message_batches/create_message_batches/test_success.py
beoh1: I don't mind it. I think it make it clear the value is coming from the input data and keeps it simple.

TheWorldWideWebb (approved):

PR #396, opened 2024-01-08
CCM-2787: Update sandbox responses

Kennedy1265:
## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 1 reviews received

### DIFF
--- a/tests/lib/assertions.py
+++ b/tests/lib/assertions.py
@@ -30,6 +30,8 @@ class Assertions():
         assert resp.status_code == 200
 
         response = resp.json().get("data")
+        messageStatus = response.get("attributes").get("messageStatus")
+        
         assert response.get("type") == "Message"
         assert response.get("id") is not None
         assert response.get("id") != ""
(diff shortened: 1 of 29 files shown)


tests/lib/assertions.py
```diff
@@ -39,9 +41,35 @@ def assert_200_response_message(resp, environment):
...
+        if messageStatus != "pending_enrichment":
+            assert response.get("attributes").get("metadata") is not None
+            assert response.get("attributes").get("metadata") != ""
+            assert response.get("attributes").get("metadata")[0].get("queriedAt") is not None
+            assert response.get("attributes").get("metadata")[0].get("queriedAt") != ""
+            assert response.get("attributes").get("metadata")[0].get("source") is not None
+            assert response.get("attributes").get("metadata")[0].get("source") != ""
+            assert response.get("attributes").get("metadata")[0].get("version") is not None
+            assert response.get("attributes").get("metadata")[0].get("version") != ""
+            assert response.get("attributes").get("metadata")[0].get("labels") != ""
```
adadigital, line 60: I noticed you're checking each value `is not None` or `!= ""` but on labels you're not doing the check for None, is that deliberate? And potentially do we want to check for an empty list here?

I like the directness of these assertions but I want to check they're doing what you think they're doing. `dict.get` is handy for dictionary lookups which you expect could fail but it will, by default, return None if that item was not found. Therefore if `"attributes"` isn't a key in `response` you'll be calling `None.get("metadata")[0].get("labels")`, which will throw an `AttributeError` since `None` has no such method `get`. That's fine if that's what you want! It just seems funny to me to use `get` like you're programming it defensively but exposing yourself to that error anyway.

You could do `response.get("attributes",{}).get("metadata",[{}])` if you were aiming to get `None` out as an output if any part of the chain failed. Or you could assert that that chain exists before making assertions against it if that's getting a bit complicated and boilerplate-y?

Interested in your thoughts but trust your judgement regardless.

adadigital (approved):
I spent some time trying to figure out how you made the decisions about what the correct `labels` field was for each of the sandbox messages but wasn't quite sure. Would be interested in where that information came from.

tests/lib/assertions.py
Kennedy1265: On labels in the sandbox responses - I did an analysis of the differences between dev and sandbox and realised each instance where there was a channel present, i.e: sending, delivered, failed due to missing whatever etc there was also a label for which channel was targeted. I didn't get this from a design document anywhere so I was making an assumption based on behaviour I'd noticed in dev.

On `assert_200_response_message` as a whole its serving each response possible with a 200 status, which unfortunately means that getting specific can be a bit tricky - the not null or not empty checks are there to make sure that its populated with _something_ but we do lose out on checking explicitly the body of separate types of response. The labels check not having `is not None` was written (not written?) to allow for the responses return empty arrays which would fail if this were added. I like the idea of specifically testing that something would be empty so could definitely use  your `response.get("attributes",{}).get("metadata",[{}]` in a potential future refactor where we split out the tests and check for expected content

Honestly I went with this because its relatively light weight (wouldn't have to do a refactor of assertions) and served multiple requests without issue, though I wouldn't be opposed to writing something that would specifically cater to the values we'd expect back.

Kennedy1265: These are some good points to address, but may be a little out of scope for this ticket so I've raised https://nhsd-jira.digital.nhs.uk/browse/CCM-2808 to address these points

PR #399, opened 2024-01-11
WIP: CCM-2141 429s to include default response headers

adadigital:
## Summary

This moves the CORS headers into a separate file and includes them from two places (to save duplicating them) so that the 429 response also includes the same headers. We found we had to set `IgnoreUnresolvedVariables` to `true` in the 429 fault to allow apigee to handle an unspecified origin as it did before, otherwise the API throws a 500 internal server error when it would otherwise throw a 429 but no origin is specified.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 1 reviews received

### DIFF
--- a/proxies/shared/policies/Quota.Global.xml
+++ b/proxies/shared/policies/Quota.Global.xml
@@ -12,7 +12,7 @@ For more information on rate limiting, see the following resources:
     <DisplayName>Quota.Global</DisplayName>
     <Interval>1</Interval>
     <TimeUnit>minute</TimeUnit>
-    <Allow count="6000"/>
+    <Allow count="5"/>
     <AsynchronousConfiguration>
         <SyncMessageCount>100</SyncMessageCount>
     </AsynchronousConfiguration>
--- a/proxies/shared/policies/RaiseFault.429TooManyRequests.xml
+++ b/proxies/shared/policies/RaiseFault.429TooManyRequests.xml
@@ -45,5 +45,5 @@
             </Payload>
         </Set>
     </FaultResponse>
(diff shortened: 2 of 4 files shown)


proxies/shared/policies/Quota.Global.xml
```diff
@@ -12,7 +12,7 @@ For more information on rate limiting, see the following resources:
     <DisplayName>Quota.Global</DisplayName>
     <Interval>1</Interval>
     <TimeUnit>minute</TimeUnit>
-    <Allow count="6000"/>
```
Ian-Hodges, line 15: Was this meant to be a permanent change?

proxies/shared/policies/RaiseFault.429TooManyRequests.xml
```diff
@@ -17,9 +17,12 @@
         <Set>
             {% if ENVIRONMENT_TYPE == 'sandbox' %}
```
Ian-Hodges, line 18: Small suggestion. Slightly more succinct: 
```suggestion
          <Headers>
                  [% include './partials/Partial.Headers.CorsHeaders.xml' %]
					        {% if ENVIRONMENT_TYPE == 'sandbox' %}
					        <Header name="Retry-After">5</Header>
					        {% endif %}
          </Headers>
          <StatusCode>429</StatusCode>
```

proxies/shared/policies/Quota.Global.xml
adadigital: Nope, will be removed, just holding this here because I don't know if Phil approves of the approach we took yet

PR #420, opened 2024-01-30
CCM-1692: Email end to end tests

Kennedy1265:
## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


Makefile
```diff
@@ -193,3 +193,12 @@ mtls-test:
 	--ignore=tests/integration \
```
Ian-Hodges, line 193: With the addition of the test/end_to_end folder, do we need extra ignore flags here and above?
Is there an `include` flag? It would be easier to manage than all these `ignore` flags.

Kennedy1265: I did look at include for a bit but it appears that that is related to using multiple Makefiles so didn't seem appropriate for this purpose, I did however, find that  when running tests locally it appears that declaring the file path the tests are under the test command seems to do the job without the need for adding in loads of ignores.

Ian-Hodges: Your solution is far neater than mine.

Ian-Hodges (approved):

PR #437, opened 2024-02-12
CCM-1693: Add letter e2e tests

Kennedy1265:
## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval

### DIFF
--- a/tests/lib/assertions.py
+++ b/tests/lib/assertions.py
@@ -159,17 +159,34 @@ class Assertions():
         assert actual == expected
 
     @staticmethod
-    def assert_message_delivered_gov_uk(response, message_id, type):
-        for i in range(len(response)):
-            if message_id in response[i].get("reference"):
-                assert response[i].get("status") == "delivered"
-                assert response[i].get("type") == type
-                assert response[i].get("body") is not None
-                if type == "email":
-                    assert response[i].get("subject") is not None
-                    assert response[i].get("email_address") is not None
-                if type == "sms":
-                    assert response[i].get("phone_number") is not None
+    def assert_email_gov_uk(response, message_id):
+        for message in response:
+            if message_id in message.get("reference"):
+                assert message.get("status") == "delivered"
+                assert message.get("type") == "email"
+                assert message.get("body") is not None
+                assert message.get("subject") is not None
+                assert message.get("email_address") is not None
+                break
+
+    @staticmethod
+    def assert_sms_gov_uk(response, message_id):
+        for message in response:
+            if message_id in message.get("reference"):
+                assert message.get("status") == "delivered"
+                assert message.get("type") == "sms"
+                assert message.get("body") is not None
+                assert message.get("phone_number") is not None
+                break
+
+    @staticmethod
+    def assert_letter_gov_uk(response, message_id):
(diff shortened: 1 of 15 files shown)


tests/lib/assertions.py
```diff
@@ -159,19 +159,34 @@ def assert_200_valid_message_id_response_body(resp, message_id, url):
...
+            if message_id in response[i].get("reference"):
+                assert response[i].get("status") == "delivered"
+                assert response[i].get("type") == "sms"
+                assert response[i].get("body") is not None
+                assert response[i].get("phone_number") is not None
                 break
 
+    @staticmethod
+    def assert_letter_gov_uk(response, message_id):
+        for i in range(len(response)):
```
adadigital, line 183: I appreciate you're just copying a pattern and if you're content with copying this pattern no problemo, but just checking you are aware that indexable objects in python are also often iterable, so you can just do `for message in response:` and then `assert message.get("status") == "received"` and so on and so forth. Might look a bit nicer?

tests/lib/helper.py
```diff
@@ -0,0 +1,39 @@
...
+                    "Accept": "application/vnd.api+json"
+                },
+            )
+
+            if get_message_response.status_code == 200:
+                message_status = get_message_response.json().get("data").get("attributes").get("messageStatus")
+            time.sleep(10)
+
+        if message_status != end_state:
+            raise TimeoutError(f"Request took too long to be processed. Message id: {message_id}")
```
adadigital, line 39: Maybe record `message_status` in the TimeoutError so that it's easier to see why something failed? (i.e. sending vs failed)

Also is it worth considering any terminal states in this function and terminating early if they're reached? (such as failed) Would save it from taking 5 minutes for every type of problem.

tests/lib/assertions.py
```diff
@@ -159,19 +159,34 @@ def assert_200_valid_message_id_response_body(resp, message_id, url):
...
-                assert response[i].get("type") == type
+                assert response[i].get("type") == "email"
                 assert response[i].get("body") is not None
-                if type == "email":
-                    assert response[i].get("subject") is not None
-                    assert response[i].get("email_address") is not None
-                if type == "sms":
-                    assert response[i].get("phone_number") is not None
+                assert response[i].get("subject") is not None
+                assert response[i].get("email_address") is not None
```
adadigital, line 169: Could also drop a break on the end of here (and the one for letter) to save it from iterating through everything, I'm not sure how large this response could potentially be or if it's a problem if we don't.

adadigital (approved):
Just wanted to say I think it's awesome that we're moving those calls to the API into a really simple "send single message" helper function.

tests/lib/assertions.py
Kennedy1265: Looks a load better, cheers for the suggestion!

tests/lib/helper.py
Kennedy1265: Added a check in the while for failed status to break out if it was encountered so that should trim down run time in that eventuality- also added the message status as well. good shout

PR #503, opened 2024-04-05
CCM-3970: create user details endpoint

aaron-downing:
## Summary

User details endpoint with the following available responses:
- 200
  - Supply T00001 for paginated responses (8 static pages available)
  - All other valid ODS codes and X26 code have 1 static page available by default
- 400
  - Invalid ODS code
  - ODS code not provided.
- 404
  - Report not found

Following responses are to be retrieved using the `Prefer` header with a value of `code=429`/`code=500` when sending to the real APIM proxy endpoint.
- 429
  - Too many requests (send `rate_limit` in the Authorization header)
- 500
  - Something went wrong (send `server_error` in the Authorization header)

Also had to change the sonar scanner docker image to 5.0 instead of 5.0.0 as it was causing the static analysis job to fail - seems to be an issue related to Node 18.7 becoming deprecated - quick fix was outlined in this thread https://community.sonarsource.com/t/node-18-17-deprecated/112524/5

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


sandbox/handlers/user_details.js
```diff
@@ -0,0 +1,102 @@
...
+      429,
+      "Too many requests"
+    );
+    next();
+    return;
+  }
+
+  const odsCode = req.params.odsCode.toUpperCase()
+
+  if (!odsCodeRegex.test(odsCode) && odsCode !== 'X26') {
```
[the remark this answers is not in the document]
aaron-downing: It is the ODS code that can be used to retrieve all patients registered within the app - I think X26 represents NHS Digital as a whole

simonlabarere (approved):

sandbox/handlers/user_details.js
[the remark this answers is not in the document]
adadigital: Also wondering why we test that something matches a regular expression but then immediately enforce that it matches X26? Am I correct in understanding that this is saying "if the ODS code doesn't match the regular expression and the ODS code is not X26 then throw a 400 error"?

TheWorldWideWebb (approved):

sandbox/handlers/user_details.js
```diff
@@ -0,0 +1,102 @@
...
+    sendError(
+      res,
+      500,
+      "Something went wrong"
+    );
+    next();
+    return;
+  }
+
+  if (req.headers["authorization"] === "rate_limit") {
```
adadigital, line 28: I'm not sure we actually implement 429s in the sandbox application itself, I think the APIM code has cases for, when running in the sandbox, it looks at the "prefer" header and returns that response if included for any endpoint, hence why this isn't implemented for any other endpoint in the sandbox.

Would it be worth checking the behaviour with this code removed? This might also apply to some other error codes like a 500

aaron-downing: I shall give this a go locally, thank you for the pointer

adadigital: When you push your code to the MR it should automatically deploy to APIM and you can at least pass requests through with the sandbox environment you deploy, do you know how to find the URL?

aaron-downing: Removing ability to serve 429s and 500s from the sandbox code and instead relying on the APIM route of using the `Prefer` header with `code=500`/`code=429` in a request to get these responses

sandbox/__test__/user_details.spec.js
```diff
@@ -0,0 +1,215 @@
...
+    });
+
+    after(function () {
+        process.env = env;
+        server.close();
+    });
+
+    it("returns a service ban (403) when the user is banned", (done) => {
+        request(server)
+            .get("/comms/channels/app/supplier/nhsapp/accounts")
```
simonlabarere, line 21: This is the path on the APIM side, on sandbox I believe it should look more like `/api/channels/app/supplier/nhsapp/accounts` to reflect what we will call on API Gateway.

aaron-downing: Updated

simonlabarere (approved):

simonlabarere (approved):

TheWorldWideWebb (approved):
Looking nice with the new %PATH_ROOT% links now

PR #511, opened 2024-04-10
CCM-4070: speed up zap scan

TheWorldWideWebb:
### Description

SInce the addition of the get messages endpoint the duration of the zap scan job has gone from 25 minutes to around 2.5 hours. The max timeout (without paying) of an azure job is 60 minutes. The following changes were made to decrease scan duration.

1. Stopped the scanning of all three endpoints when it is only supposed to scan one.
2. Decreased the delay from 1000 to 500 milliseconds.
3. Only run rules appropriate for an API.

The API policy rules were copied over from the [configuration file](https://www.zaproxy.org/docs/docker/api-scan/#configuration-file) used for ZAP's own api scan script.

### Test Evidence

A new scan completed in 19 minutes.
[https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=204731](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=204731&view=ms.vss-test-web.build-test-results-tab)
It also reports the same vulerabilities as an old scan, before the timeout issues.
[https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=180246](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=180246&view=ms.vss-test-web.build-test-results-tab)


scripts/publish_zap_compatible.py
```diff
@@ -28,22 +28,6 @@ def scan_and_remove(obj, mappings):
 with open('build/communications-manager.json') as f:
     specification = json.load(f)
 
-# set the servers - this allows us to test unpublished environments
-specification["servers"] = [
```
TheWorldWideWebb, line 32: Does the removal of this actually render the openapi spec generated for Zap empty ? 

The communications-manager-zap.json ?

If so and this now still works, can we remove that code all together ?

tdroza-nhs (approved):

simonlabarere (approved):

PR #548, opened 2024-04-29
CCM-4566 fix zap vulnerabilities

Ian-Hodges:
## Summary

Actioning the ZAP issues so the workflow runs green.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [x] 2 reviews received
* [x] Tester approval

## Test Evidence

Successful security scan: https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=211413&view=results

#### 10049 - Non-Storable Content
We do not want the responses to be cached, and set the headers to indicate this.
image

#### 90005 - Sec-Fetch-Site Header is Missing
Sec-Fetch-* headers are only for requests. This rule is not relevant for APIs on their own.

#### 10094 - Base64 Disclosure
The KSUIDs are sometimes being detected as base64 strings.
image

#### 110009 - Full Path Disclosure
In our error responses we include a link to the NHS developer catalogue.
image
The ZAP code is picking up `developer`.
image

#### 90004 - Insufficient Site Isolation Against Spectre Vulnerability
When there is no Origin header in the request, we do not include the CORS headers in the response. This includes the Cross-Origin-Resource-Policy, which is what ZAP is flagging.

#### 10062 - PII Disclosure
In our error responses we include the apigee message id in the body.
image
Sometimes it can be detected as a Maestro credit card number. Note, this rule only checks for credit card details, no other PII.
image

### Removal of Production Scans
These ZAP scans are not intended to be ran against a production environment. Unintended actions could be performed. The Integration environment is sufficiently production-like to test against.

### CORS Headers
The CORS headers were not implemented correctly. It has been changed so an OPTIONS request returns the following CORS headers.
With an origin:
image
Without:
image
All other requests return the following headers.
With an origin:
image
Without:
image

### Remove Scanning of /callbacks 
There is a callback endpoint in the api specifications, but it's not one we host. It is there as a guide for consumers on setting up such an endpoint on their system to receive status updates from NHS Notify. Including it has no benefit and increases the scan duration.

### DIFF
--- a/scripts/publish_zap_compatible.py
+++ b/scripts/publish_zap_compatible.py
@@ -34,7 +34,8 @@ with open('build/communications-manager-zap.json', 'w') as f:
             specification,
             (
                 ("format", "date"),
-                ("personalisation", None)
+                ("personalisation", None),
+                ("/callbacks", None)
             )
         ),
         f
--- a/zap/zap.yaml
+++ b/zap/zap.yaml
@@ -10,41 +10,20 @@ env:
       urls:
         - https://int.api.service.nhs.uk/comms # the actual url used for the scan is defined in the openapi job below
       authentication:
-        method: 'script'
+        method: script
         parameters:
           url: https://int.api.service.nhs.uk/oauth2/token
-          script: 'scripts/authentication/get_bearer_token.js'
-          scriptEngine: 'Graal.js'
+          script: scripts/authentication/get_bearer_token.js
+          scriptEngine: Graal.js
         verification:
-          method: 'response'
-          loggedOutRegex: '401 Unauthorized'
+          method: response
+          loggedOutRegex: 401 Unauthorized
       users:
         - name: Integration
           credentials:
             kid: local
-            api_key: "${INTEGRATION_API_KEY}"
-            private_key: "${INTEGRATION_PRIVATE_KEY_CONTENTS}"
-    - name: ProdUnauthenticated
-      urls:
-        - https://api.service.nhs.uk/comms # the actual url used for the scan is defined in the openapi job below
-    - name: ProdAuthenticated
-      urls:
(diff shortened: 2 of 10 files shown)


scripts/publish_zap_compatible.py
```diff
@@ -34,7 +34,8 @@ def scan_and_remove(obj, mappings):
             specification,
             (
                 ("format", "date"),
-                ("personalisation", None)
+                ("personalisation", None),
+                ("/callbacks", None)
```
aaron-downing, line 38: Is this to stop us scanning the /callbacks endpoints we have defined in our spec?

aaron-downing: Lol just scrolled down and saw you've added a description of the changes on the PR, nice work

zap/zap.yaml
```diff
@@ -54,46 +33,63 @@ jobs:
...
+      engine: Graal.js
+      name: get_bearer_token
+      file: scripts/authentication/get_bearer_token.js
+
+  # configure the passive scan
+  - type: passiveScan-config
+    parameters:
+      maxAlertsPerRule: 10
+      scanOnlyInScope: true
+    rules:
```
tdroza-nhs, line 54: Could you add a quick one-line comment to each of these rules to summarise why they have been disabled so that the next person that sees it doesn't try to re-add them?

aaron-downing (approved):

TheWorldWideWebb (approved):

zap/zap.yaml
Ian-Hodges: Comments added.

tdroza-nhs (approved):

PR #586, opened 2024-05-29
CCM-2615: added NHS Number link to docs in spec

DavidFueryBJSS:
## Summary
Added NHSnumber link info to status 400 info in:
https://digital.nhs.uk/developer/api-catalogue/nhs-notify#post-/v1/message-batches
https://digital.nhs.uk/developer/api-catalogue/nhs-notify#post-/v1/messages

## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [x] 2 reviews received
* [x] Tester approval

### DIFF
--- a/specification/schemas/types/LinksErrorNHSNumber.yaml
+++ b/specification/schemas/types/LinksErrorNHSNumber.yaml
@@ -6,4 +6,9 @@ properties:
     format: uri
     example: 'https://digital.nhs.uk/developer/api-catalogue/communications-manager'
     description: 'https://digital.nhs.uk/developer/api-catalogue/communications-manager'
+  nhsNumbers:
+    type: string
+    format: uri
+    example: 'https://www.datadictionary.nhs.uk/attributes/nhs_number.html'
+    description: https://www.datadictionary.nhs.uk/attributes/nhs_number.html This link is only provided in the instance of a "CM_INVALID_NHS_NUMBER" error.
 minProperties: 1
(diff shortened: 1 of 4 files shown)


specification/schemas/types/LinksError.yaml
```diff
@@ -3,8 +3,7 @@ additionalProperties: false
 properties:
   about:
     type: string
-    enum:
-      - 'https://digital.nhs.uk/developer/api-catalogue/communications-manager'
     format: uri
     example: 'https://digital.nhs.uk/developer/api-catalogue/communications-manager'
```
tdroza-nhs, line 7: Although there's currently a redirect in place, can you update these to the new nhs-notify url please? e.g. `https://digital.nhs.uk/developer/api-catalogue/nhs-notify`

specification/schemas/types/LinksErrorNHSNumber.yaml
```diff
@@ -0,0 +1,14 @@
+type: object
+additionalProperties: false
+properties:
+  about:
+    type: string
+    format: uri
+    example: 'https://digital.nhs.uk/developer/api-catalogue/communications-manager'
```
tdroza-nhs, line 7: As above re `communications-manager` -> `nhs-notify` urls

specification/schemas/types/LinksError.yaml
tdroza-nhs: Actually, given that this is an example of a response message that we return, can you check whether we've updated the urls in our actual responses yet - if we haven't then it's fine to leave these pointing to the old urls

adadigital (approved):

specification/schemas/types/LinksError.yaml
tdroza-nhs: Investigated by David - The API response still returns the old `communications-manager` path so the docs are correct that they match this. The urls will be updated by CCM-3068

PR #596, opened 2024-06-07
CCM-3799: documentation for nhsapp account details endpoint

adadigital:
## Summary

This merge request completes the documentation for the new NHS App account details endpoint

Main text:
maintext

### 2xx

200 example:
200example

200 schema:
200schema

### 4xx

400 example:
400example

400 schema:
400schema

401 example:
401example

401 schema:
401schema

403 example:
403example

403 schema:
403schema

404 example:
404example

404 schema:
404schema

429 example:
429example

429 schema:
429schema

### 5xx

500 example:
500example

500 schema:
500schema

502 example:
502example

502 schema:
502schema

503 example:
503example

503 schema:
503schema

504 example:
504example

504 schema:
504schema

### Amendments

We updated the documentation to include the query path parameters, and the documentation of the id field in the response

Capture

Capture2

We updated the 429 response

Capture
Capture2

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] ~Branch deployed to a dynamic env - link to deployment pipeline~
* [ ] 2 reviews received
* [ ] Tester approval


specification/responses/5xx/502_BadGateway.yaml
```diff
@@ -0,0 +1,21 @@
+description: |+
+  NHS Notify is currently unable to communicate with a downstream API required to process your request.
```
tdroza-nhs, line 2: I think the references in the 502 and 504 responses should refer to "upstream" services rather than "downstream" because a downstream service depends on an upstream service, e.g. front-end depends on back-end, back-end depends on NHSApp etc so front-end is downstream of back-end  and back-end is downstream of NHSApp.

Or a better reasoned argument from W3C:
> The terms "upstream" and "downstream" are used to describe directional requirements in relation to the message flow: all messages flow from upstream to downstream.

and:
> The 504 (Gateway Timeout) status code indicates that the server, while acting as a gateway or proxy, did not receive a timely response from an upstream server it needed to access in order to complete the request.

adadigital: Downstream was chosen to reflect the language used in the examples I received from Simon, which in turn I believe was used to reflect the language used by the live documentation. If it is your intention to use the language upstream, would you like me to change the other parts of the documentation that currently use the language upstream and communicate to Simon and Terry that we'll be needing to change the responses that the API gives?

Capture3

adadigital: We discussed this on a call and agreed to keep it as downstream to avoid code changes to 502 and 504 responses for now

tdroza-nhs (approved):

specification/responses/4xx/get_nhsapp_accounts/429_GetNhsAppAccountTooManyRequests.yaml
```diff
@@ -0,0 +1,23 @@
+description: |+
+  You have made too many requests too quickly, you must send requests at a slower rate.
```
tdroza-nhs, line 2: I think we should reword this - given the nature of this endpoint the rate limiting happens to the combined total of all requests across all clients so the client that receives the 429 may not have made too many requests. E.g. ClientA sends 500 requests in 60seconds, then ClientB sends 1 request at the end of that 60 seconds window. ClientB would also get a 429.

I'd suggest replacing this with the line from the 429 payload: "This endpoint is currently receiving a high volume of requests and is being rate limited"

adadigital: Updated

tdroza-nhs (approved):

tdroza-nhs (approved):
Re-reviewed and approved with latest commits.

PR #624, opened 2024-06-24
CCM-5115 using fixtures for authentication

adadigital:
## Summary

This changes tests to use fixtures to provide bearer tokens. The Authentication class has been changed to an AuthenticationCache that gets created once per test session. For tests in PR environments we also now add the environment to the key in the `Comms-manager-local` apigee application. This allows us to choose explicitly what key and what application a test uses.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [x] 2 reviews received
* [ ] Tester approval


simonlabarere (approved):

tests/end_to_end/test_letter.py
```diff
@@ -1,62 +1,65 @@
...
+def test_letter_end_to_end_uat(nhsd_apim_proxy_url, bearer_token_internal_dev):
     """
     .. include:: ../../partials/happy_path/test_letter_end_to_end_uat.rst
     """
     resp = Helper.send_single_message(
         nhsd_apim_proxy_url,
-        nhsd_apim_auth_headers,
-        Generators.generate_send_message_body("letter", "internal-qa")
+        {"Authorization": bearer_token_internal_dev},
+        Generators.generate_send_message_body("letter", "internal-qa"),
```
aaron-downing, line 47: Minor, but I realise this `bearer_token_internal_dev` is being used for all the non production tests would it not be better named `bearer_token_non_prod`?  I'm presuming it's this token that provides access to the comms_manager_local app in APIGEE

tests/lib/app_keys.py
```diff
@@ -0,0 +1,39 @@
+
+def assert_api_product_in_application(developer_app_keys_api, dev_email, app_name, key_id, api_product_name):
+    r = developer_app_keys_api.get_app_key(
```
aaron-downing, line 3: what is `r`?

tests/lib/authentication.py
```diff
@@ -28,7 +30,7 @@ def generate_authentication(env):
         else:
             raise ValueError("Unknown value: ", env)
 
-        if tokens.get(env, None) is None or (call_time is None or call_time + 595 < int(time())):
+        if env not in self.tokens or self.tokens[env][1] + 595 < int(time()):
```
aaron-downing, line 33: Can we make a local var of `self.tokens[env][1]` and annotate what it is? I'm still unsure of what call_time was meant to denote previously

tests/lib/authentication.py
```diff
@@ -56,7 +58,6 @@ def generate_authentication(env):
             )
             details = json.loads(resp.content)
 
-            tokens[env] = (f"Bearer {details.get('access_token')}")
-            call_time = int(time())
+            self.tokens[env] = (f"Bearer {details.get('access_token')}", int(time()))
 
-        return tokens.get(env)
+        return self.tokens[env][0]
```
aaron-downing, line 63: Again what is this value?

aaron-downing: I'm presuming the value at tokens[env] is some kind of object containing different values?

tests/lib/app_keys.py
```diff
@@ -0,0 +1,39 @@
+
+def assert_api_product_in_application(developer_app_keys_api, dev_email, app_name, key_id, api_product_name):
```
aaron-downing, line 2: Not too keen on the term `assert` here but I can't come up with a name for what this method is doing. It's more along the lines of `add_api_product_if_not_exist_in_application`

tests/end_to_end/test_letter.py
adadigital: In terms of what the fixture is called I'm not too fussy but this is liable to change again when we add a second apigee client. How would we differentiate between the two nonprod tokens that we might have? Is int prod or nonprod? By prod do we mean live environments or path to live environments and what is in our path to live?

`internal-dev` at least sticks to the convention we're using in `AuthenticationCache` whereby it calls `.generate_authentication('internal-dev')`, seems weirder to me that we'd be calling a fixture called `bearer_token_nonprod` which in turn gets its token by calling `authentication_cache.generate_authentication('internal-dev')`. At least this way it's consistent.

If you disagree I'm happy to rename it to what you want since it's not too deep a decision and it's just a search and replace job at this point. I just want us to be very firm on what we want because it's referenced in so many places that it's a pain if we keep going back and forth on this.

tests/lib/app_keys.py
adadigital: It's just a variable to denote whatever response we get back from `.get_app_key`, it's short here because the scope is very short so it's easy to see where it's being used. i.e. It's not likely you'll have to ever scroll up to see what `r` is because this function almost always fits on one screen. I can change it to `response` if needed, but this doesn't do a lot to help a reader understand the shape of that variable. One still has to go and look at the docs for `.get_app_key`, which in turn is not very helpful and takes you to the apigee docs which are a nightmare.

One could change this to annotate what we're pulling out fo the response better. And call it `current_api_products_in_app_key = developer_app_keys_api.get_app_key(...)['apiProducts']`. I thought it would be handy for this function to return its response should somebody want to use other details from that response. Shall I implement this suggestion and sacrifice the full returned body?

tests/lib/authentication.py
adadigital: `call_time` was the unix timestamp when the `Authentication` class last retrieved a (read: any) token. This meant that if I was using `Authentication` to manage multiple tokens it would actually renew each one more frequently than necessary.

My change moves `call_time` into the second index of a tuple stored in `self.tokens`, so each token records when it was issued. I'll annotate this as you've requested.

tests/lib/authentication.py
adadigital: That's the bearer token defined just two lines up:

```
self.tokens[env] = (f"Bearer {details.get('access_token')}", int(time()))
```

tests/end_to_end/test_letter.py
aaron-downing: No problem, upon realising `authentication_cache.generate_authentication('internal-dev')` had been called all along and you've just created the wrapper on top of that, we can leave it be for now

tests/lib/authentication.py
aaron-downing: Does it not make sense to name it along the lines of `token_creation_time`, `call_time` is possibly a bit ambiguous

tests/lib/authentication.py
aaron-downing: face palm

tests/lib/authentication.py
```diff
@@ -56,7 +58,6 @@ def generate_authentication(env):
             )
             details = json.loads(resp.content)
 
-            tokens[env] = (f"Bearer {details.get('access_token')}")
-            call_time = int(time())
+            self.tokens[env] = (f"Bearer {details.get('access_token')}", int(time()))
```
aaron-downing, line 61: It would be nice to give context as to what `int(time())` is meant to represent by creating a local var like was done previously (e.g. `call_time = int(time())`), but as mentioned I'm not sure I like `call_time`

adadigital: It's tied to the specific token, I don't really like the approach of using a global variable like it did before. I've put this in a tuple but if that's really an issue I think we can use namespaces or annotated tuples or dictionaries. Just seemed like overkill given it's so self contained.

tests/lib/app_keys.py
adadigital: My expectation is that when this function is called it will either return having added the api product to the given application key or it will have thrown an exception. Even though we do not explicitly throw an exception the underlying library call will throw an exception if there is a problem. How about `ensure_api_product_in_application`?

tests/lib/authentication.py
adadigital: Is the current implementation better?

tests/lib/app_keys.py
aaron-downing: `ensure_api_product_in_application` sounds good

aaron-downing (approved):

PR #625, opened 2024-06-24
CCM-5469: updated response body for 400 errors in send message and send batch docs

DavidFueryBJSS:
## Summary
updated response body for 400 errors in send message and send batch docs

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [x] 2 reviews received
* [x] Tester approval

### DIFF
--- a/specification/schemas/enums/ErrorInvalidNHSNumber.yaml
+++ b/specification/schemas/enums/ErrorInvalidNHSNumber.yaml
@@ -8,5 +8,5 @@ enum:
   - CM_TOO_FEW_ITEMS
   - CM_ODS_CODE_REQUIRED
   - CM_CANNOT_SET_ODS_CODE
-example: CM_INVALID_VALUE
-title: Enum_Error_InvalidValue
+example: CM_INVALID_NHS_NUMBER
+title: Enum_Error_InvalidNHSNumber
--- a/specification/schemas/responses/errors/message_batches/UnableToProcessBatch.yaml
+++ b/specification/schemas/responses/errors/message_batches/UnableToProcessBatch.yaml
@@ -13,7 +13,7 @@ properties:
         id:
           $ref: ../../../types/ErrorIdentifier.yaml
         code:
-          $ref: ../../../enums/ErrorInvalidValue.yaml
+          $ref: ../../../enums/ErrorInvalidNHSNumber.yaml
         links:
           $ref: ../../../types/LinksErrorNHSNumber.yaml
         status:
@@ -29,7 +29,8 @@ properties:
             - Invalid value
             - Duplicate value
             - Too few items
-          example: Missing property
+            - Invalid nhs number
+          example: Invalid nhs number
         detail:
           type: string
           enum:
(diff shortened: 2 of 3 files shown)


specification/schemas/enums/ErrorInvalidNHSNumber.yaml
```diff
@@ -0,0 +1,12 @@
+type: string
```
TheWorldWideWebb, line 1: This file name seems wrong for this content, if you look at all the error files, theyre single enums. This is more like NHS Channel Statuses.

TheWorldWideWebb (changes requested):

simonlabarere:
It looks like you updated the specs but not the actual behaviour of the API Proxy. Doesn't this need more work? Happy to help / sorry if I'm missing the plot :)

specification/schemas/enums/ErrorInvalidNHSNumber.yaml
DavidFueryBJSS: This file name is just copying the pattern used for the file it was duplicated from that has almost identical content

TheWorldWideWebb:
@simonlabarere As a part of CCM-2615. The response body for a 400 request was updated to include an error message reporting an Invalid NHS Number.

simonlabarere:
> @simonlabarere As a part of CCM-2615. The response body for a 400 request was updated to include an error message reporting an Invalid NHS Number.

Ahh that makes more sense... Thank you

specification/schemas/enums/ErrorInvalidNHSNumber.yaml
TheWorldWideWebb: I found it 

`type: string
enum:
  - CM_MISSING_VALUE
  - CM_NULL_VALUE
  - CM_INVALID_VALUE
  - CM_INVALID_NHS_NUMBER
  - CM_DUPLICATE_VALUE
  - CM_TOO_FEW_ITEMS
  - CM_ODS_CODE_REQUIRED
  - CM_CANNOT_SET_ODS_CODE
example: CM_INVALID_NHS_NUMBER
title: Enum_Error_InvalidValue`


my bad 😄

TheWorldWideWebb (approved):

specification/schemas/enums/ErrorInvalidNHSNumber.yaml
simonlabarere: I'm confused by this as well, why make a copy of `ErrorInvalidValue.yaml` and rename it to `ErrorInvalidNHSNumber.yaml` given that only one of these errors is related to the NHS Number?

simonlabarere: Sorry for unresolving this, I swear I'm usually not picky during reviews!

I don't understand why those errors should be under "Invalid NHS Number"

DavidFueryBJSS: It's specifically down to the example field on the enum - in case the `ErrorInvalidValue.yaml` is used elsewhere I've duplicated it to have the example as NHS Number in this file, if we understood the documentation framework better there might be a neat way to do this, but that would probably be a pretty deep dive to find out.

DavidFueryBJSS: This would also raise the question of why is the `ErrorInvalideValue.yaml` file named as such when it contains all definitions for all errors and not just the InvalidValue error?

simonlabarere: Happy to resolve but unless I am mistaken the only difference with with `ErrorInvalidValue.yaml` and `ErrorInvalidNHSNumber.yaml` is their title, and it doesn't make sense to me to group all those errors as "Invalid NHS Number" because that's just no true.

simonlabarere (approved):

danh-nhs (approved):
Tested on local and confirmed changes

PR #626, opened 2024-06-24
CCM-4969: update test assertions for auto generated correlation ids

TheWorldWideWebb:
as part of CCM-4969, the apigee config was changed to autogenerate correlation ids for all requests and to map them back in the response from api gateway. This broke one of the test assertions which checked correlation id was not present if not provided. 

I have updated this to check for an autogenerated id starting with rrt for apigee generated ids.

## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval

### DIFF
--- a/tests/lib/assertions.py
+++ b/tests/lib/assertions.py
@@ -271,11 +271,12 @@ class Assertions():
             else:
                 assert error in response_errors
 
-        resCorrelationID = resp.headers.get("X-Correlation-Id")
+        res_correlation_id = resp.headers.get("X-Correlation-Id")
         # apigee generates this value if not present with rrt prefix
-        isValidCorrelationID = (resCorrelationID == correlation_id
-                                or (correlation_id is None and resCorrelationID.startswith('rrt')))
-        assert isValidCorrelationID
+        if correlation_id:
+            assert res_correlation_id == correlation_id
+        else:
+            assert res_correlation_id.startswith('rrt')
 
         # ensure we have our x-content-type-options set correctly
         assert resp.headers.get("X-Content-Type-Options") == "nosniff"
--- a/tests/lib/assertions.py
+++ b/tests/lib/assertions.py
@@ -271,7 +271,7 @@ class Assertions():
             else:
                 assert error in response_errors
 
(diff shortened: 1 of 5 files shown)


tests/lib/assertions.py
```diff
@@ -271,7 +271,11 @@ def assert_error_with_optional_correlation_id(resp, code, error, correlation_id)
             else:
                 assert error in response_errors
 
-        assert resp.headers.get("X-Correlation-Id") == correlation_id
+        resCorrelationID = resp.headers.get("X-Correlation-Id")
+        # apigee generates this value if not present with rrt prefix
+        isValidCorrelationID = (resCorrelationID == correlation_id
+                                or  (correlation_id is None and resCorrelationID.startswith('rrt')))
+        assert isValidCorrelationID
```
adadigital, line 278: pytest uses assertion rewrites to basically tell you what went wrong in a pretty format (and we in fact benefited from customising our assertion representation function, see `tests/conftest.py`), if this fails it's going to go "AssertionError: False", and it might be a bit tricky to debug? What you've put will work, but it might be a pain later down the line.

If I offer you the counter suggestion..
```
if correlation_id:
    assert res_correlation_id == correlation_id
else:
    assert res_correlation_id.startswith('rrt')
```

Would that seem to you like an upgrade or a downgrade? Let me know, and I'll trust your call.

Other than that, hard limit: we don't use camel case in python. I'm not strict to the point of using a linter but camel case is for classes in PEP8 and for everything else we separate words with underscores. Hoping we can self regulate this rather than getting to the point of adding another linter to worry about.

adadigital (changes requested):

tests/lib/assertions.py
adadigital: This is what I mean by PEP8 if you're interested: https://peps.python.org/pep-0008/

adadigital (approved):

simonlabarere:
Now part of https://github.com/NHSDigital/communications-manager-api/pull/628

PR #675, opened 2024-07-19
CCM-5954 linter

adadigital:
## Summary

This change brings in a way to run a linter for the sandbox application by invoking `node_modules/.bin/eslint sandbox/`, and makes a best effort attempt at resolving issues flagged by the linter in this way.

The linter rules are based on what we implemented in [comms-mgr](https://github.com/NHSDigital/comms-mgr), but we need to drop the linter rule about not permitting `.js` extensions in import statements because base javascript does not support importing from files without the full filename.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


sandbox/handlers/batch_send.js
```diff
@@ -76,20 +75,21 @@ export async function batch_send(req, res, next) {
...
   // Note: the docker container uses node:12 which does not support optional chaining
-  const odsCodes = messages.map((message) => {
-    if (message && message.originator) {
-      return message.originator.odsCode;
-    }
-  });
-  if (odsCodes.includes(undefined) && req.headers["authorization"] === noDefaultOdsClientAuth) {
+  const odsCodes = messages.map((message) => message && message.originator ? message.originator.odsCode : undefined
+  );
+//  const odsCodes = messages.map((message) => {
```
lapenna-bjss, line 87: did you mean to leave this in?

adadigital: No, removed, thankyou

sandbox/app.js
```diff
@@ -1,14 +1,11 @@
-"use strict";
```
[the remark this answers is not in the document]
simonlabarere: Looks like that's fixed :)

simonlabarere (approved):

sandbox/app.js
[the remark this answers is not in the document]
lapenna-bjss: Yes, that was a good catch, thank you! :)

sandbox/handlers/get_message.js
```diff
@@ -1,18 +1,19 @@
...
+  if (req.headers.authorization === "banned") {
     sendError(res, 403, "Request rejected because client service ban is in effect");
     next();
     return;
   }
 
-  const messageId = req.params.messageId;
+  const { messageId } = req.params;
 
+  /* eslint-disable-next-line security/detect-non-literal-fs-filename */
```
adadigital, line 13: Are we sure we don't want to resolve these rather than just put in an ignore statement? It's trying to highlight to us that one could come along and go `curl http://sandbox.url/messages?messageId=../../../../../some-important-json-file.json`

Maybe this doesn't matter since it's only our sandbox app which is only intended for internal use, but in that case I would suggest we should disable the linter rule entirely rather than just put in an ignore statement?

lapenna-bjss: Yes, I didn't think it mattered much since it's only the sandbox app, but that's a good point. I'll disable the rule for sandox.

tdroza-nhs (approved):

PR #680, opened 2024-07-23
CCM-5754: Fail zap pipeline when error in bash

simonlabarere:
## Summary
Updates to the Zap scan:
- Update bash script to exit on a failure. The zap scan is run inside a docker container and if it failed then the script would go on and the pipeline would still show as green.
- Update the openapi specs used by zap in order to exclude the callback URLs defined in the documentation. The 4 endpoints scanned are now: Get Message, Post Message, Post Batch and Get NHS App Accounts.
- Update the definition of the SupplierStatus and ChannelStatus enums to have a unique file name. This was causing a warning followed by an automatic appending of the suffix "-2"
- Use the previous version of zap: 2.14.0. The current version 2.15.0 is causing some errors that I believe are from the jython script `zap/scripts/authentication/get_bearer_token.js` but I couldn't get to the bottom of them. https://nhsd-jira.digital.nhs.uk/browse/CCM-5987 raised to look into it.

This is now the output of the scan:
image

Note that the errors `[Fatal Error] :1:1: Content is not allowed in prolog.` are a red herring and can be ignored. Each scan is only run once.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [x] 2 reviews received
* [x] Tester approval

### DIFF
--- a/scripts/run_zap.sh
+++ b/scripts/run_zap.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-set -o nounset errexit pipefail
-set -e
+set -o nounset
+set -o errexit
+set -o pipefail
 
 # create our temporary directory
 export TEMP_DIR=/tmp
--- a/scripts/run_zap.sh
+++ b/scripts/run_zap.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -o nounset errexit pipefail
+set -e
 
 # create our temporary directory
 export TEMP_DIR=/tmp
--- a/azure/nightly-zap-scan-pipeline.yml
+++ b/azure/nightly-zap-scan-pipeline.yml
@@ -41,7 +41,6 @@ steps:
     parameters:
       webhook_uri: 'ALERTS_DEV_API_WEBHOOK_URI'
   - bash: |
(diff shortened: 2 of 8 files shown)


scripts/run_zap.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -o nounset errexit pipefail
+set -e
```
Ian-Hodges, line 4: This shouldn't be needed. `errexit` on line 3 above is the equivalent of `set -e`.

simonlabarere: From my testing it didn't work when the bash job run in the docker container failed, I can show you if you have a minute and see what you think.

simonlabarere: Scratch that you're absolutely right.
Updating those statements to the following seems to do the trick.
```bash
set -o nounset
set -o errexit
set -o pipefail
```

Ian-Hodges (approved):

lapenna-bjss (approved):

Kennedy1265 (approved):

PR #706, opened 2024-08-20
CCM-5025: Contact details override error responses

simonlabarere:
## Summary
This change is to handle an array of errors returned by the backend. Each error item contains:
- A title
- A pointer to the field causing the error
- A message

The implementation has been kept generic for now (no hardcoded status codes or reason phrases) in the hope that it can be reused in the future.

Note that automated tests are supposed to be added as part of [CCM-5328](https://nhsd-jira.digital.nhs.uk/browse/CCM-5328).

#### Create Message
image

#### Create Batch
image

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval

### DIFF
--- a/proxies/shared/partials/Partial.Target.FaultRules.xml
+++ b/proxies/shared/partials/Partial.Target.FaultRules.xml
@@ -1,7 +1,7 @@
 <FaultRule name="combined_backend_fault_handler">
     [% include './partials/Partial.Component.SetResponseDefaults.xml' %]
     <Step>
-        <Name>ExtractVariables.NhsAppAccounts.Error</Name>
+        <Name>ExtractVariables.ErrorMessage</Name>
         <Condition>
             response.content ~~ "\{\u0022message\u0022:\u0022.*\u0022\}"
         </Condition>
--- a/proxies/shared/partials/Partial.Target.FaultRules.xml
+++ b/proxies/shared/partials/Partial.Target.FaultRules.xml
@@ -7,6 +7,24 @@
         </Condition>
     </Step>
     <Step>
+        <Name>ExtractVariables.ErrorDetails</Name>
+        <Condition>
+            response.content ~~ ".*\u0022errors\u0022:.*"
+        </Condition>
+    </Step>
+    <Step>
+        <Name>JavaScript.EnhanceErrorDetails</Name>
+        <Condition>
+            data.errors != null
+        </Condition>
+    </Step>
+    <Step>
+        <Name>RaiseFault.GenericErrorDetails</Name>
+        <Condition>
+            data.enhancedErrors != null
+        </Condition>
+    </Step>
+    <Step>
         <Name>RaiseFault.404NhsAppAccounts</Name>
         <Condition>
             (proxy.pathsuffix MatchesPath "/channels/nhsapp/accounts") and response.status.code = 404 and data.errorMessage != null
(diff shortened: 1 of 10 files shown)


proxies/shared/partials/Partial.Target.FaultRules.xml
```diff
@@ -6,6 +6,24 @@
             response.content ~~ "\{\u0022message\u0022:\u0022.*\u0022\}"
         </Condition>
     </Step>
+    <Step>
```
aaron-downing, line 9: Not too clued up on APIGEE side of things, but do these steps get applied to all errors returned from the API irregardless of the endpoint?

sandbox/handlers/messages.js
```diff
@@ -23,6 +24,16 @@ function getOriginatorOdsCode(req) {
   return odsCode;
 }
 
+function getEmailOverride(req) {
```
aaron-downing, line 27: This should avoid any null reference error
```suggestion
function getEmailOverride(req) {
 return req?.body?.data?.attributes?.recipient?.contactDetails?.email
}
```

sandbox/handlers/messages.js
```diff
@@ -119,6 +130,20 @@ export async function messages(req, res, next) {
     return;
   }
 
+  const emailOverride = getEmailOverride(req);
```
aaron-downing, line 133: Unit test

proxies/shared/partials/Partial.Target.FaultRules.xml
simonlabarere: Correct, I personally don't think we should try and restrict these new policies for a given set of endpoints. Keeping those errors generic might be helpful in the future.

sandbox/handlers/messages.js
simonlabarere: I copied an existing function which was probably meant for an older version of node. I'll make the change, thank you.

simonlabarere: (This got marked as resolved automatically)

proxies/shared/partials/Partial.Target.FaultRules.xml
aaron-downing: Are they applied sequentially? I'm just wondering if there's any issues with applying the 1st step to a none NhsAppAccounts endpoint error? (I realise this is in prod already and I'm lacking a lot of knowledge in this area)

simonlabarere: Yes they apply sequentially. I first step extract the message from the response from the BE and puts it in a context variable. It's a bit of a mess at the moment, it was added as part of the NHS App Accounts work but I don't think that policy should have "NhsAppAccounts" in its name.

aaron-downing: Oh ok, I mean if you're happy to, could rename it to something more generic?

simonlabarere: Done :)

aaron-downing (approved):

aaron-downing:
Awesome work 🎆

lapenna-bjss (approved):

PR #711, opened 2024-08-28
CCM-5327: add contact detail override and validation to sandbox

aaron-downing:
## Summary

Introduced ability to provide contact detail overrides to the single message and batch message endpoints with following validation rules:
- `sms` (optional)
  - Must be `string`
- `email` (optional)
  - Must be `string`
- `address` (optional)
  - Must be an `object`
  - Must contain `lines` field which must be an `Array` that has at least 2 entries but less than 6. Each value must be a `string`
  - Must contain `postcode` field which must be a string

Authorization header must have the following value to use contact detail override: `allowedContactDetailOverride`

It's possible to trigger invalid phone number and invalid email address errors by using the following values:
- sms: `"07700900002"`
- email: `"invalidEmailAddress"`

Additionally aligned common validation rules between the single message and batch message endpoints. Single message endpoint is now checking that
- Request body contains `data` attribute
- Request body contains `data.attributes` attribute
- Request body contains `data.type` attribute that equals `'Message'`
- Request body contains `data.attributes.routingPlanId` attribute

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


simonlabarere (approved):
Nice work but I personally think that sandbox app is becoming more complex than it needs to be. Effectively we only needed 1 or 2 validation errors with the new `errors` array just to make sure it was translated as expected by the API layer. Mocking all these errors here is a bit overkill in my humble opinion.

aaron-downing:
> Nice work but I personally think that that sandbox app is becoming more complex than it needs to be. Effectively we only needed 1 or 2 validation errors with the new `errors` array just to make sure it was translated as expected by the API layer. Mocking all these errors here is a bit overkill in my humble opinion.

Thank you. True though I wanted to ensure that there was correct API schema validation in place but ignore any business logic related to what is a valid value e.g. email addresses, postcodes, phone numbers.

lapenna-bjss (approved):

PR #720, opened 2024-09-10
CCM-5786: Fix schema validation errors caused by invalid type definitions

tdroza-nhs:
## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval

### DIFF
--- a/specification/communications-manager.yaml
+++ b/specification/communications-manager.yaml
@@ -33,20 +33,7 @@ paths:
     get:
       $ref: endpoints/get_message.yaml
   /channels/nhsapp/accounts:
-    parameters:
-      - schema:
-          $ref: schemas/types/ODSOrganisationCode.yaml
-        name: ods-organisation-code
-        in: query
-        required: true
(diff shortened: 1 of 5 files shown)


specification/communications-manager.yaml
```diff
@@ -33,20 +33,7 @@ paths:
     get:
       $ref: endpoints/get_message.yaml
   /channels/nhsapp/accounts:
-    parameters:
```
tdroza-nhs, line 36: These are both duplicated under the `get:` endpoint below so aren't required here

specification/schemas/components/MessageStatus.yaml
```diff
@@ -2,7 +2,10 @@ type: object
 title: CallbackRequest
 properties:
   type:
-    const: MessageStatus
```
tdroza-nhs, line 5: `const` is not a valid OpenAPI schema keyword, so I've replaced it with a string that has a single value to replicate the same behaviour

aaron-downing (approved):

danh-nhs (approved):

PR #763, opened 2024-10-10
CCM-6648: Fix intermittent 401 errors

simonlabarere:
## Summary

Edit: Unfortunately we still got 401 errors after implementing the solution below because we can get 100s of successful tests before getting random failures. I think we can still keep this solution but I also added retries for individual tests.

It can take up to 5 minutes for Apigee to register that a Product has been added to an Application. This can sometimes cause APIM tests to fail because of 401 errors (and tests that are expecting 401s to be false positives).

Some logic has been added to the token generation for tests in order to test the token multiple times to make sure the authentication is working before proceeding to the tests. If the authentication is not ready (the test request returns a 401) then the logic waits 10 seconds before trying again.

One would think the correct place for this would be where the product is added to the application but an auth token is not easily available from there and that piece of code is currently independent from the environment it's running on, whereas auth tokens do depend on the environment. I think it also makes sense to test a token at the time it's generated.

Token Generation drawio

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval

### DIFF
--- a/tests/lib/assertions.py
+++ b/tests/lib/assertions.py
@@ -243,15 +243,7 @@ class Assertions():
 
     @staticmethod
     def assert_error_with_optional_correlation_id(resp, code, error, correlation_id):
-        if code == 429:
-            # If we are testing a 429 error then only retry on 504 errors
-            error_handler.handle_504_retry(resp)
-        elif code == 504:
-            # If we are testing a 504 error then only retry on 429 errors
-            error_handler.handle_429_retry(resp)
-        else:
-            # We are not testing a 429 or 504, retry if we get either of them
-            error_handler.handle_retry(resp)
+        error_handler.handle_retry(resp, excluded_code=code)
 
         assert resp.status_code == code, f"Response: {resp.status_code}: {resp.text}"
 
--- a/tests/lib/assertions.py
+++ b/tests/lib/assertions.py
@@ -243,7 +243,9 @@ class Assertions():
 
     @staticmethod
     def assert_error_with_optional_correlation_id(resp, code, error, correlation_id):
-        error_handler.handle_retry(resp, excluded_code=code)
+        if resp.status_code != code:
+            # Unexpected status code, check if the test needs to be retried
+            error_handler.handle_retry(resp)
 
         assert resp.status_code == code, f"Response: {resp.status_code}: {resp.text}"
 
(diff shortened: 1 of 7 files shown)


tests/lib/fixtures.py
```diff
@@ -73,23 +74,25 @@ def authentication_cache():
...
+def bearer_token_internal_dev(authentication_cache, api_product_in_comms_manager_local, nhsd_apim_proxy_url):
+    return authentication_cache.generate_authentication('internal-dev', nhsd_apim_proxy_url)
 
 
 @pytest.fixture
-def bearer_token_internal_dev_test_1(authentication_cache, api_product_in_nhs_notify_test_client_1):
-    return authentication_cache.generate_authentication('internal-dev-test-1')
+def bearer_token_internal_dev_test_1(authentication_cache,
+                                     api_product_in_nhs_notify_test_client_1,
+                                     nhsd_apim_proxy_url):
```
aaron-downing, line 84: This call signature has been updated but it doesn't look like you've updated anywhere that it's used. Can this method be removed or is there some smart parameter passing going on when this method gets used?

simonlabarere: This fixture is used by rate limiting tests: https://github.com/NHSDigital/communications-manager-api/blob/fix/CCM-6648_Intermittent_401_errors/tests/development/rate_limiting/test_429_errors.py

The fixture is injected as a test parameter and pytest will build it using this function.

Happy to explain on a call!

aaron-downing: Ah neat, I thought it might be something clever under the hood

aaron-downing (approved):

DavidFueryBJSS (approved):

tests/lib/assertions.py
```diff
@@ -243,15 +243,7 @@ def assert_messages_idempotency(resp_one, resp_two):
 
     @staticmethod
     def assert_error_with_optional_correlation_id(resp, code, error, correlation_id):
-        if code == 429:
-            # If we are testing a 429 error then only retry on 504 errors
-            error_handler.handle_504_retry(resp)
-        elif code == 504:
-            # If we are testing a 504 error then only retry on 429 errors
-            error_handler.handle_429_retry(resp)
```
aaron-downing, line 251: Are these individual handle X retry methods now redundant or are they being used elsewhere?

aaron-downing (approved):

PR #849, opened 2025-01-03
CCM-8114: Remove parametrisation from tests using NHS Notify Test Client

Kennedy1265:
## Summary
This PR resolves the intermittent test failure on internal-dev for the test 'test_not_permitted_to_use_contact_details'. The intermittent failure was caused by two tests attempting to set a rate limit value on the same client at the same time.

Removing the correlation-id aspect of the test ensures that only one test is ran using NHS Notify Client ID during the standard internal-dev test run so we can continue to run the whole test suite with 4 parallel runners without significantly compromising on the value of what is being tested.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


simonlabarere (approved):
I think another solution would have been to make those tests use different APIM clients. Although I assume that as a tester you're happy with the removal of the correlation ID!

lapenna-bjss (approved):

PR #878, opened 2025-02-17
CCM-8460: Update tests to use one suite for all environments (excluding sandbox)

Kennedy1265:
## Summary
* Created a fixture to provide a url that would point to a given environment based off the value provided in API_ENVIRONMENT and API Product name
* Created a fixture to provide authentication for a given environment based off the value provided in API_ENVIRONMENT
* Created a directory that would house the new test suite
* Added a simple test for get message that creates a message and verifies the GET message endpoint returns a 200.
* Added a 400 test to be in line with existing test suites
* Updated pipeline yml to set environment variables API_ENVIRONMENT and PROXY_NAME for all environments so it would be compatible with the new test command
* Added a "test" command to the Makefile and pytest.ini to run the new test suite with this argument 
* Amended generators to use a routing config that work across all environments and updated the default NHS number to use the production smoke test user  on generate_valid_create_message_body
* Added a default headers function to generators
* Tested that the new test command would work on [internal-dev](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=294990&view=results), [internal-qa](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=294970&view=results), [int](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=294974&view=results) and [prod](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=294976&view=results) nightly jobs and in the [internal-dev pr pipeline](https://dev.azure.com/NHSD-APIM/API%20Platform/_build/results?buildId=294994&view=results) job. *Note* this has since been reverted and will be fully implemented in [CCM-8633](https://nhsd-jira.digital.nhs.uk/browse/CCM-8633)

## Key decisions:
The environment variable API_ENVIRONMENT was chosen as the determining factor in what environment a test was running on because that value is set by the 'environment' parameter passed into the pipeline yml for each pipeline job and would always be present when running tests. amendments were made to run-pipelines.yml to get the new tests working on each environment in cases where API_ENVIRONMENT was not exported as an environment variable. There was also an issue with PROXY_NAME not being recognised in the int and prod environments when running in the pipeline so credentials were provided to get tests to pass. 

The decision to remake the get message endpoint test rather than re-use an existing test was made because the tests currently use inconsistently seeded data across all environments. creating a message and then verifying that message can be retrieved via the GET message endpoint verifies the system works as expected in a more "end to end" manner although with this change there is some loss in validating the expected response body as requests transition through different stages of processing. Also it should be noted that there is no delay between sending a message and retrieving its status so it is likely that the retrieved message will often be stuck in a state of "Pending enrichment".

One of the acceptance criteria in [CCM-8460](https://nhsd-jira.digital.nhs.uk/browse/CCM-8460) was to check if fixtures can be moved to conftest.py rather than its own folder, I was able to verify that this was possible but the change would have altered most of the files present across all test suites, and as they are about to be refactored into a singular test suite it would be more effective to implement this at a later stage, a ticket has been raised to address this [CCM-8632](https://nhsd-jira.digital.nhs.uk/browse/CCM-8632)

Another of the acceptance criteria in [CCM-8460](https://nhsd-jira.digital.nhs.uk/browse/CCM-8460) was to check if it was possible to include sandbox tests as part of this work, but as the sandbox uses stock responses, functionality like creating a message and verifying it has been created via the get message endpoint would not work. Also, with the sandbox suite separated we could perhaps use sandbox as a place to test the expected responses of a request in more detail but use the internal-dev, int and prod suites to check that the service works as intended across each environments, allowing us to cut out some of the unnecessary bulk in the tests targeting these environments.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


Makefile
```diff
@@ -212,3 +212,17 @@ e2e-test-uat:
 	$(TEST_CMD) \
 	tests/end_to_end \
 	-m "e2e and uattest"
+
+.environment:
+	@if [ -n "$(API_ENVIRONMENT)" ]; then \
+		printf "\n\nTests are running on $(API_ENVIRONMENT)\n\n\n"; \
+	else \
+		printf "\nMissing environment variables\n\n"; exit 1; \
```
tdroza-nhs, line 220: Might be useful if the error included the name of the missing environment variable?

pytest.ini
```diff
@@ -12,4 +12,5 @@ markers =
     devperftest: rate limiting tests suitable to run against dev like environments
     inttest: suitable to run against integration environment
     prodtest: suitable to run against production environment
-    uattest: suitable to run against uat environment
\ No newline at end of file
+    uattest: suitable to run against uat environment
+    test: suitableto run against all environments
```
tdroza-nhs, line 16:
```suggestion
    test: suitable to run against all environments
```

tdroza-nhs (changes requested):

tdroza-nhs (approved):

adadigital (approved):

PR #955, opened 2025-06-16
CCM-10553: Fix Dependabot Redoc 

ScottFullerton-NHSE:
## Summary
I've run pnpm update -r to update the files to see if this will fix the dependabot issue. However, it looks to have updated possibly more then we originally wanted. I've put this PR here to see what others think before attempting to merge it and see if it clears.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,30 +25,30 @@
     ]
   },
   "devDependencies": {
-    "@babel/core": "^7.25.2",
-    "@babel/eslint-parser": "^7.25.1",
+    "@babel/core": "^7.27.4",
+    "@babel/eslint-parser": "^7.27.5",
     "@redocly/cli": "^1.34.3",
     "eslint": "^8.57.1",
     "eslint-config-airbnb-base": "^15.0.0",
-    "eslint-config-prettier": "^8.5.0",
+    "eslint-config-prettier": "^8.10.0",
     "eslint-formatter-gitlab": "^3.0.0",
     "eslint-plugin-html": "^6.2.0",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-jsx-a11y": "^6.6.1",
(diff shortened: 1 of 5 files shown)


tdroza-nhs (changes requested):
There a lot of change here but all-but-one appear to be minor or patch version updates that I'm not too concerned about. However, the one that worries me is ajv as that's 4 major version increments!
```
    "ajv": "4.11.8 - 8",
    "ajv": "^8.17.1",
```

I think it might be safer to back that one out of this PR and apply it separately under another ticket so that if it doesn't cause issues we can roll it back without reverting all of the other updates

package.json
```diff
@@ -25,30 +25,30 @@
...
-    "newman": "^6.1.2",
+    "minimist": "^1.2.8",
+    "newman": "^6.2.1",
     "sinon": "^17.0.1",
-    "typescript": "^5.5.4"
+    "typescript": "^5.8.3"
   },
   "dependencies": {
-    "ajv": "4.11.8 - 8",
+    "ajv": "4.11.8",
```
tdroza-nhs, line 51: I think this needs to go back to the following because you've now fixed it at 4.11.8 where previously it was allowing any version between 4.11.8 and version 8 so some of the dependencies in pnpm-lock.yaml have now been downgraded from 6.x down to 4.11.8
```suggestion
    "ajv": "4.11.8 - 8",
```

tdroza-nhs (changes requested):

simonlabarere (approved):

simonlabarere (changes requested):
We need to investigate the sandbox unit test failures before merging this.

simonlabarere (approved):

PR #1068, opened 2026-05-20
CCM-17830: Double PROD rate limits

lapenna-bjss:
## Summary

The global API rate limits configured for production are intermittently being breached by SMS Nudge traffic. This PR doubles the current quota and spike arrest values

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


manifest_template.yml
```diff
@@ -46,6 +46,8 @@ APIGEE_ENVIRONMENTS:
     - name: prod
       display_name: Production
       approval_type: manual
+      global_quota: '12000'
```
markpullan: Can we change the defaults so that production is considered the default setup, and non-prod environments are a conscious deviation away from our prod config?

(Achieves the same goal, but aligns better to the philosophy of treating production as the standard and every deviation from prod as the special case, rather than prod itself being a special case)

markpullan (changes requested):

manifest_template.yml
gareth-allan: This should now be done @markpullan.

markpullan (approved):

simonlabarere (approved):

PR #1090, opened 2026-08-04
CCM-22090: Recipient response message retrieval API

rhyscoxnhs:
## Summary
Adds proxy configuration for new recipient response message retrieval API.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Checklist
* [ ] 2 reviews received
* [ ] Tester approval


proxies/shared/policies/ExtractVariables.MessageResponses.Get.Request.xml
```diff
@@ -0,0 +1,33 @@
...
+    </QueryParam>
+    <QueryParam name="pageSize">
+        <Pattern ignoreCase="true">{pageSize}</Pattern>
+    </QueryParam>
+    <QueryParam name="authoredAtFrom">
+        <Pattern ignoreCase="true">{authoredAtFrom}</Pattern>
+    </QueryParam>
+    <QueryParam name="authoredAtTo">
+        <Pattern ignoreCase="true">{authoredAtTo}</Pattern>
+    </QueryParam>
```
cgitim: I don't think we need any of these.

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
```diff
@@ -0,0 +1,33 @@
...
+        <Verb>GET</Verb>
+        <QueryParams>
+            <QueryParam name="responseId">{data.responseId}</QueryParam>
+            <QueryParam name="clientId">{data.clientId}</QueryParam>
+            <QueryParam name="correlationId">{data.correlationId}</QueryParam>
+            <QueryParam name="cursor">{data.cursor}</QueryParam>
+            <QueryParam name="pageSize">{data.pageSize}</QueryParam>
+            <QueryParam name="authoredAtFrom">{data.authoredAtFrom}</QueryParam>
+            <QueryParam name="authoredAtTo">{data.authoredAtTo}</QueryParam>
+        </QueryParams>
```
cgitim: Don't need query params

sandbox/handlers/message_responses.js
```diff
@@ -0,0 +1,67 @@
...
+  return {
+    messageId,
+    responses: [
+      {
+        responseId: '22222222-2222-4222-8222-222222222222',
+        messageReference: 'msg-ref-1',
+        code: 'YES',
+        channel: 'nhsapp',
+        channelStatus: 'delivered',
+        authoredAt: '2026-01-02T09:00:00.000Z'
```
cgitim: I think we need `cascadeType` in here as well.

Plus, `messageId`.

specification/documentation/APIDescription.md
```diff
@@ -255,6 +255,10 @@ In order to present the recipient with answers, include the `answerOptions` fiel
 
 If you subscribe to recipient response callbacks, NHS Notify will send you a callback when a recipient responds to a message (currently only NHS App supports this). See [the recipient response callback](#post-/-client-provided-recipient-response-URI-) for more details.
 
+### Retrieving responses via the API
+
+Use the [get message responses](#get-/v1/messages/-messageId-/responses) endpoint to query historical responses for a given message. Responses are available for up to 9 months after the message was sent.
```
cgitim: This URL is wrong

specification/documentation/GetResponses.md
```diff
@@ -0,0 +1,217 @@
+## Overview
+
+Use this endpoint to query historical NHS App keyword responses associated with a specific message.
```
cgitim: Probably better to make generic (i.e., just recipient responses, not NHS App keyword responses)

specification/documentation/GetResponses.md
```diff
@@ -0,0 +1,217 @@
...
+### Authentication
+
+This endpoint requires APIM application identity credentials. You must include a valid `Authorization` header on every request. See the [security and authorisation](#overview--security-and-authorisation) section for details on obtaining credentials.
+
+### Preconditions
+
+* Your application has valid APIM credentials with the appropriate scope.
+* The message used the NHS App channel with keyword response options configured.
+
+If no responses exist yet, the `data` array in the response will be empty.
```
cgitim: Don't need to mention authentication

specification/documentation/GetResponses.md
```diff
@@ -0,0 +1,217 @@
...
+| `code=503` | 503 Service unavailable |
+
+Example:
+
+```
+curl -X GET \
+  --header "Accept: application/vnd.api+json" \
+  --header "Prefer: code=403" \
+  https://sandbox.api.service.nhs.uk/comms/v1/messages/2WL3qFTEFM0qMY8xjRbt1LIKCzM/responses
+```
```
cgitim: This whole file departs significantly from the other API calls... See `CreateMessage.md` and `GetMessage.md`.

specification/endpoints/get_responses.yaml
```diff
@@ -0,0 +1,34 @@
+summary: Get responses for a message
+description:
+  $ref: ../documentation/GetResponses.md
```
cgitim: Should probably be `../documentation/GetMessageResponses.md`?

specification/schemas/components/ResponseItem.yaml
```diff
@@ -0,0 +1,42 @@
+type: object
+title: ResponseItem
+additionalProperties: false
+required:
+  - responseId
+  - authoredAt
+  - channel
+  - channelStatus
+  - code
+  - messageReference
```
cgitim: Include
* `cascadeType`
* `messageId`

specification/responses/2xx/200_Responses.yaml
```diff
@@ -0,0 +1,10 @@
+description: The responses for the given message have been retrieved successfully.
+headers:
+  $ref: ../../snippets/StandardResponseHeaders.yaml
+content:
+  application/vnd.api+json:
+    schema:
+      $ref: ../../schemas/responses/GetResponsesSuccess.yaml
```
cgitim: `GetResponsesSuccess.yaml` should probably be renamed for consistency.

`MessageResponsesResponse.yaml` would be consistent, but ugly.

If we're using "responses" as an abbreviation for message responses, then `ResponsesResponse.yaml` would also be consistent.

specification/communications-manager.yaml
```diff
@@ -32,6 +32,16 @@ paths:
         description: The unique identifier for the message.
     get:
       $ref: endpoints/get_message.yaml
+  /v1/messages/{messageId}/responses:
```
cgitim: Wrong URL

sandbox/handlers/message_responses.js
```diff
@@ -0,0 +1,67 @@
...
+    next();
+    return;
+  }
+
+  res.type('json').status(200).json(getDefaultResponse(messageId));
+}
+
+function getDefaultResponse(messageId) {
+  return {
+    messageId,
```
cgitim: I think we're removing this?

specification/schemas/responses/GetResponsesSuccess.yaml
```diff
@@ -0,0 +1,13 @@
+title: GetResponsesSuccess
+type: object
+additionalProperties: false
+required:
+  - messageId
+  - responses
+properties:
+  messageId:
+    $ref: ../types/KSUID.yaml
```
cgitim: Remove `messageId`, we moved to response items

cgitim (changes requested):

proxies/shared/policies/ExtractVariables.MessageResponses.Get.Request.xml
rhyscoxnhs: Fixed

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
rhyscoxnhs: Fixed

sandbox/handlers/message_responses.js
rhyscoxnhs: Fixed

specification/documentation/APIDescription.md
rhyscoxnhs: Fixed

specification/documentation/GetResponses.md
rhyscoxnhs: Fixed

specification/documentation/GetResponses.md
rhyscoxnhs: Fixed

specification/documentation/GetResponses.md
rhyscoxnhs: Fixed

specification/endpoints/get_responses.yaml
rhyscoxnhs: Fixed

specification/schemas/components/ResponseItem.yaml
rhyscoxnhs: Fixed

specification/responses/2xx/200_Responses.yaml
rhyscoxnhs: Fixed

specification/communications-manager.yaml
rhyscoxnhs: Fixed

sandbox/handlers/message_responses.js
rhyscoxnhs: Fixed

specification/schemas/responses/GetResponsesSuccess.yaml
rhyscoxnhs: Fixed

specification/documentation/GetMessageResponses.md
```diff
@@ -0,0 +1,41 @@
+## Overview
+
+Use this endpoint to retrieve recipient responses associated with a specific message.
```
mjewildnhs, line 3: From the PR comments I can see we changed this to "recipient responses".
I don't think it would be immediately clear what this is.
They are introduced a bit in the "Two-way messaging" section above.
But the Recipient Response section (which is lower on the page) has a "About recipient responses" section which provides a bit of context on what they are.
Maybe at the very least we should have a "see Two-way messaging" link or something so the reader can figure out what they are.

specification/documentation/GetMessageResponses.md
```diff
@@ -0,0 +1,41 @@
...
+
+Responses are available for retrieval for up to 9 months after the message was sent.
+
+### Response structure
+
+A successful response returns a flat JSON array of response items.
```
mjewildnhs:
```suggestion
A successful response returns an array of response items.
```
For readability
Don't think JSON is needed and flat definitely not.

specification/responses/4xx/message_responses/422_TooManyResponses.yaml
```diff
@@ -0,0 +1,24 @@
+description: |+
+  There are too many responses associated with this message to return in a single response. This error occurs when more than 1000 responses exist for the given message ID.
```
mjewildnhs: The problem with documenting this scenario is we really need some guidance on what to do in this scenario. I don't know what that guidance is though! It's raise a support query with us / NHS App but I don't think we want to document that.
My first question as a dev writing code to implement response handling is what do I do with this response code.
Baking it into the sandbox makes you think its something you'll need to deal with and test but the reality is it shouldn't ever happen.
I think we are better changing the code to return a 500 so the generic 500 documentation catches it.

specification/endpoints/get_responses.yaml
```diff
@@ -0,0 +1,34 @@
...
+    $ref: ../responses/4xx/message_responses/400_UnableToGetResponses.yaml
+  '401':
+    $ref: ../responses/4xx/401_AccessDenied.yaml
+  '403':
+    $ref: ../responses/4xx/403_Forbidden.yaml
+  '404':
+    $ref: ../responses/4xx/404_NotFound.yaml
+  '405':
+    $ref: ../responses/4xx/405_NotAllowed.yaml
+  '406':
```
mjewildnhs: The accepted values documented for a 406 are:
```  
  * `*/*`
  * `application/json`
  * `application/vnd.api+json`
  * `application/json; charset=utf-8`
  * `application/vnd.api+json; charset=utf-8`
```
I think `application/json` is the only 1 we accept.
Despite this I'm not sure it will return a 406 and we need to test it.

A related problem is 415
We know from UAT that API gateway does return a 415 if the content type isn't `application/json`
We don't have 415 as a possibility in the docs.

We need to decide on which variants we should support.
AI seems to be under the belief `vnd.api+json` isn't the right one:
"The payload described here is a plain JSON array of response objects, not a JSON:API document. 
A JSON:API response would normally be wrapped in a data envelope and use JSON:API conventions such as links, meta, and resource-object structure."

The other documented responses do use a data wrapper.
Which suggest we should be.

I think we need to:
- determine what happens for an unsupported accept type - AI is suggesting it won't be a 406
- document the 415
- decide we need to use the data wrapper
- add support for `application/vnd.api+json` et al

If we don't add `vnd.api` support then we need to remove all mention of it as its in the curl examples etc

specification/responses/4xx/message_responses/400_UnableToGetResponses.yaml
```diff
@@ -0,0 +1,30 @@
+description: |+
+  The message ID in the request path is not a valid KSUID.
+
+  The following validation error can occur:
+
+  | Error code | Title | Description |
+  | ---------- | ----- | ----------- |
+  | `CM_INVALID_VALUE` | Invalid value | The `messageId` path parameter is not a valid KSUID. |
```
mjewildnhs: Where's `CM_INVALID_VALUE` come from the app-response code returns "Invalid messageId format"
I'm a bit unsure on the error handling in general with this APIM stuff it looks like error responses are an array of ids,codes,etc - none of which we are doing in app-response.
I'm unsure where all these `CM` codes are defined.

specification/schemas/enums/ErrorTooManyResponses.yaml
mjewildnhs, line 1: Is `CM_TOO_MANY_RESPONSES` served/handled by APIM?

sandbox/handlers/message_responses.js
mjewildnhs, line 1: This doesn't seem to line up with the docs.
It doesn't handle all codes e.g. 429
It handles undocumented codes e.g. 502
It also returns error messages rather than error codes (though I'm not sure where those codes are supposed to come from and I think the documentation or our implementation needs to change).
As per my other comment I think we need a thorough review of these error responses.

specification/documentation/APIDescription.md
```diff
@@ -255,6 +255,10 @@ In order to present the recipient with answers, include the `answerOptions` fiel
 
 If you subscribe to recipient response callbacks, NHS Notify will send you a callback when a recipient responds to a message (currently only NHS App supports this). See [the recipient response callback](#post-/-client-provided-recipient-response-URI-) for more details.
 
+### Retrieving responses via the API
+
+Use the [get message responses](#get-/v1/message-responses/-messageId-) endpoint to query historical responses for a given message. Responses are available for up to 9 months after the message was sent.
```
mjewildnhs:
```suggestion
Use the [get message responses](#get-/v1/message-responses/-messageId-) endpoint to query responses for a given message. Responses are available for up to 9 months after the message was sent.
```
I think we decided "historic" was a bit misleading?

tests/sandbox/message_responses/test_success.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+
+    assert resp.status_code == 200, f"Response: {resp.status_code}: {resp.text}"
+    body = resp.json()
+    assert body.get("messageId") == VALID_MESSAGE_ID
+    assert isinstance(body.get("responses"), list)
+    assert len(body["responses"]) > 0
+
+    first = body["responses"][0]
+    assert "responseId" in first
```
mjewildnhs, line 24: Do we want all the fields here i.e. the fields added (cascadeType, etc)?

specification/documentation/GetMessageResponses.md
rhyscoxnhs: Fixed

specification/documentation/GetMessageResponses.md
rhyscoxnhs: Fixed

specification/responses/4xx/message_responses/422_TooManyResponses.yaml
rhyscoxnhs: My personal view is if it's > HTTP 300 then it's an error, as an integrator you'd log it and go through failure handling. I'd personally treat any 4xx other than a 401 / 403 as an explicit error.

Technically that would be the case in this situation, we should never realistically have > 1000 responses, but also you could argue that this wouldn't always be our failure, it could be upstream triggering it.

I'm leaning towards a 500 being the right call, but it does imply a critical failure with our system. Is that appropriate?

specification/endpoints/get_responses.yaml
rhyscoxnhs: Fixed, documented 415, removed 406 and removed vnd.api content types.

specification/responses/4xx/message_responses/400_UnableToGetResponses.yaml
rhyscoxnhs: Fixed, we're not referencing CM codes now

specification/schemas/enums/ErrorTooManyResponses.yaml
rhyscoxnhs: Fixed, we're not referencing CM codes now

sandbox/handlers/message_responses.js
rhyscoxnhs: Fixed

specification/documentation/APIDescription.md
rhyscoxnhs: Fixed

tests/sandbox/message_responses/test_success.py
rhyscoxnhs: Fixed

specification/responses/4xx/message_responses/422_TooManyResponses.yaml
mjewildnhs: It could be a failure with our system (e.g. rogue response event is processed multiple times) or more likely a failure with the NHS App. Either way it isn't the clients fault and we probably need to take responsibility for it.
What do you think @cgitim ?

rhyscoxnhs: Yeah I think that's a good point - 4xx usually implies it's the fault of the caller of the API, whereas 5xx implies the inverse. This makes me lean much more towards a 5xx error code, although 500 still implies a crash or some critical fault in my view.

Maybe a 503 is more appropriate? It does indicate a bit more clearly that this is a temporary, resolvable situation.

specification/endpoints/get_responses.yaml
mjewildnhs, line 1: The error responses we are re-using all have CM_ prefix error codes.
I think we still need to take an decision on our deviation away from the [JSON:API](https://jsonapi.org/) convention used by the other endpoints.
Does `CM_` stand for comms-mgr?
I think its this list of error codes: https://github.com/NHSDigital/comms-mgr/blob/main/packages/libs/ingestion-lib/src/error-responses.ts#L95-L94
As it stands we are re-using the following error codes:
```
401 Access denied → CM_DENIED
403 Forbidden → CM_FORBIDDEN or CM_SERVICE_BAN
404 Not found → CM_NOT_FOUND
405 Method not allowed → CM_NOT_ALLOWED
408 Request timeout → CM_TIMEOUT
415 Unsupported media → CM_UNSUPPORTED_MEDIA
429 Too many requests → CM_QUOTA
500 Internal server error → CM_INTERNAL_SERVER_ERROR
503 Service unavailable → CM_SERVICE_UNAVAILABLE
504 Service timeout → CM_TIMEOUT
```
We'd need to check what error message and what API gateway does for some of these scenarios.

However I think for now we need to stop and have a conversation about our deviation from the JSON:API convention before making further change.
I suspect we don't want to return CM_ errors which would be a bit annoying as the CM_ errors have occupied `specification/schemas/enums`  and might need moving/scoping.

If none of the JSON:API stuff has been discussed already then I think we need some input from @cgitim

mjewildnhs (commented):
I've paused reviewing until we get an answer on whether we should be following the JSON:API convention as per my review comment.

specification/responses/4xx/message_responses/422_TooManyResponses.yaml
mjewildnhs: I don't think its that temporary or retryable. We'll need a fairly hard change at our end do something about it.
I think we did discuss it in refinement and the ticket does have "If >100 responses  are received return a 500"
https://nhsd-jira.digital.nhs.uk/browse/CCM-22090

rhyscoxnhs: We wouldn't just leave it that way though. We'd address the problem from an operational standpoint, and from that point onwards we wouldn't be returning a 5XX error code. This to me makes returning a 503 make more sense.

It is resolvable, not by the caller, but by us. It's a temporary situation, not a permanent failure. I doubt we'd leave the data in a state where the integrator would never be able to retrieve responses for that particular id again.

rhyscoxnhs: Agreed to go with 500

mark-r-bjss:
After a team discussion we decided that

1. It would be best to align the response more closely with [existing API responses](https://digital.nhs.uk/developer/api-catalogue/nhs-notify#get-/v1/messages/-messageId-) by adding in a data object wrapper and listing the items within.
2. To stick with a 500 response if there are too many responses found for the messageId as this is an unexpected real-world scenario.
3. To extract the repeated fields `messageId` and `messageReference` into the new `data` wrapper.
4. To sort the patient-responses so that the API produces consistent responses for the same query as this would make testing easier.

rhyscoxnhs:
One revision to the above - since JSON API mandates specific fields for the data wrapper, we can't extract `messageId` and `messageReference` out of the `attributes` object, the only valid fields in the data wrapper are `id`, `attributes`, `links`, `relationships` and `type`.

specification/endpoints/get_responses.yaml
rhyscoxnhs: Fixed - migrated to JSON:API

rhyscoxnhs:
Migrated to JSON:API

specification/responses/4xx/message_responses/400_MessageResponsesInvalidRequest.yaml
mjewildnhs: Missing `messageId` gives:
400 - {"message": "Missing required request parameters: [messageId]"}
Need to document here and fix in app-response

specification/endpoints/get_responses.yaml
```diff
@@ -0,0 +1,32 @@
...
+  '400':
+    $ref: ../responses/4xx/message_responses/400_MessageResponsesInvalidRequest.yaml
+  '401':
+    $ref: ../responses/4xx/401_AccessDenied.yaml
+  '403':
+    $ref: ../responses/4xx/403_Forbidden.yaml
+  '404':
+    $ref: ../responses/4xx/404_NotFound.yaml
+  '405':
+    $ref: ../responses/4xx/405_NotAllowed.yaml
```
mjewildnhs: Currently returns {"message":"Missing Authentication Token"} - think we need a reponse template for what its documented as: CM_NOT_ALLOWED

specification/documentation/GetMessageResponses.md
```diff
@@ -0,0 +1,69 @@
...
+To simulate error responses in the sandbox, use the following message IDs:
+
+* not found - `00000000-0000-4000-8000-000000000404`
+* too many responses - `00000000-0000-4000-8000-000000000500`
+
+Here's an example curl command:
+
+```
+curl -X GET 'https://sandbox.api.service.nhs.uk/comms/v1/message-responses/11111111-1111-4111-8111-111111111111' \
+     --header 'Accept: application/json'
```
mjewildnhs: Can we make sure all the curl examples use accept header "application/vnd.api+json" to align with the rest of the documented API endpoints.

specification/responses/4xx/415_UnsupportedMedia.yaml
```diff
@@ -1,16 +1,8 @@
 description: |+
-  The `Content-Type` of the request is not supported. This endpoint supports:
-
-  * `application/json`
-  * `application/vnd.api+json`
-  * `application/json; charset=utf-8`
-  * `application/vnd.api+json; charset=utf-8`
+  The `Content-Type` of the request is not supported. This endpoint supports `application/json`.
 content:
-  application/vnd.api+json:
```
mjewildnhs: Think we can use the standard 415 now we do (or at least should) support json+api

specification/endpoints/get_responses.yaml
```diff
@@ -0,0 +1,32 @@
...
+  '408':
+    $ref: ../responses/4xx/408_RequestTimeout.yaml
+  '415':
+    $ref: ../responses/4xx/415_UnsupportedMedia.yaml
+  '429':
+    $ref: ../responses/4xx/429_TooManyRequests.yaml
```
mjewildnhs: We'll return this 429 if apigee rate limits.
However if AWS API Gateway does it for some reason then I think the response will fall through the rules and return whatever the AWS Gateway response is (i.e. not json).
This is probably unlikely to happen.
However for some reason "/channels/nhsapp/accounts" have added handling for their app returning 429s
See https://github.com/NHSDigital/communications-manager-api/commit/05d0f63fe8e08100528b09d808a2246a90fc4ca3
Might be best doing the same.

sandbox/handlers/message_responses.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  if (messageId === notFoundMessageId) {
+    res.status(404).json({ error: 'No responses found for the specified messageId' });
+    next();
+    return;
+  }
+
+  if (messageId === tooManyResponsesMessageId) {
+    res.status(422).json({
```
mjewildnhs: This should be 500? 
Test for this says returns 500 but asserts 429

sandbox/handlers/message_responses.js
```diff
@@ -0,0 +1,78 @@
...
+    });
+    next();
+    return;
+  }
+
+  res.type('json').status(200).json(getDefaultResponse(messageId));
+}
+
+function getDefaultResponse(messageId) {
+  return [
```
mjewildnhs: We need to update this to match the json+api response we now return e.g. "data" field et al

sandbox/handlers/message_responses.js
```diff
@@ -0,0 +1,78 @@
+import { sendError } from './utils.js'
+
+const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+
+const notFoundMessageId = '00000000-0000-4000-8000-000000000404';
+const tooManyResponsesMessageId = '00000000-0000-4000-8000-000000000422';
+
+export async function messageResponses(req, res, next) {
+  if (req.headers.authorization === 'banned') {
+    res.status(403).json({ error: 'Forbidden' });
```
mjewildnhs: The established pattern for the sandbox is to return different response bodies to the actual documented error response...which is great.
I think we should buck the trend and do it properly (i.e. return the code, etc)

mjewildnhs (commented):
Not finished yet but posting feedback I have.

specification/responses/4xx/415_UnsupportedMedia.yaml
```diff
@@ -1,10 +1,5 @@
 description: |+
-  The `Content-Type` of the request is not supported. This endpoint supports:
```
mjewildnhs: I think we should revert this as it affects all API end points not just app-response.

tests/api/message_responses/test_200_success.py
```diff
@@ -0,0 +1,20 @@
+import requests
+import pytest
+from lib import Assertions, Generators
+from lib.constants.message_responses_paths import MESSAGE_RESPONSES_ENDPOINT, VALID_MESSAGE_ID
+from lib.fixtures import *  # NOSONAR
+
+
+@pytest.mark.devtest
```
mjewildnhs: Its a bit of a shame we can't test this in other environments.
e.g. `get_message` runs its tests in all envs
However it can create messages to then get.
We don't have a way to create a response so I guess we are stuck with dev only test which I think is against the sandbox. This does risk things breaking in prod on APIM releases.

Might be worth discussion with @cgitim

sandbox/handlers/message_responses.js
```diff
@@ -0,0 +1,138 @@
+const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+
+const notFoundMessageId = '00000000-0000-4000-8000-000000000404';
+const tooManyResponsesMessageId = '00000000-0000-4000-8000-000000000500';
+const supportedContentTypes = ['application/json', 'application/vnd.api+json'];
+
+export async function messageResponses(req, res, next) {
```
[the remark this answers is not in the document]
rhyscoxnhs: Addressed feedback - reverted to simple responses for the responses that are unconditionally overwritten, and preserved the ones that aren't.

specification/responses/4xx/415_UnsupportedMedia.yaml
rhyscoxnhs: Fixed

proxies/live/apiproxy/targets/target.xml
```diff
@@ -9,6 +9,7 @@
         [% include './partials/Partial.Flows.CreateMessageEndpoint.xml' %]
         [% include './partials/Partial.Flows.GetMessageEndpoint.xml' %]
         [% include './partials/Partial.Flows.GetNhsAppAccountsEndpoint.xml' %]
+        [% include './partials/Partial.Flows.GetMessageResponsesEndpoint.xml' %]
```
mjewildnhs: I think we'll need a new `TargetEndpoint`.
This currently routes to `comms-mgr` via the `LoadBalancer` config and we need it to route to the `app-response` bounded context.
During refinement did we discuss this and what other bounded contexts were doing. I'm sure Tim mentioned another bounded context using the proxy config but I can't see evidence of it.

rhyscoxnhs: Fixed

rhyscoxnhs: Fixed

sandbox/__test__/message_responses.spec.js
```diff
@@ -56,6 +56,13 @@ describe('/api/v1/message-responses/:messageId', () => {
             .expect('Content-Type', /json/, done);
     });
 
+    it('returns a 200 for a non-v4 UUID messageId', (done) => {
```
mjewildnhs: I got a bit confused seeing this test.
Its NON_V4 because VALID_MESSAGE_ID is v4 - which isn't version obvious
Might be clearer just iterating through all versions
```
it.each([1, 2, 3, 4, 5, 6, 7, 8])(
    'returns a 200 for a valid UUID v%s messageId',
    (version, done) => {
        const messageId = `11111111-1111-${version}111-8111-111111111111`;

        request(server)
            .get(`/api/v1/message-responses/${messageId}`)
            .expect(200)
            .expect('Content-Type', /json/, done);
    }
);
```

proxies/live/apiproxy/targets/app-response-target.xml
mjewildnhs: Do we need to update the stuff in docs?
e.g. repos/communications-manager-api/docs/proxies.md
Theres also some tests docs which might be safe to ignore for now as there might be quite a bit of churn in the test area when we actually get them up and running.

sandbox/handlers/message_responses.js
```diff
@@ -1,9 +1,9 @@
 import { sendError } from './utils.js'
 
-const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+// Mirrors the RFC 9562 UUID validation used by the app-response bounded context (versions 1-8, plus the nil/max UUIDs).
+const uuidRegex = /^([0-9a-f]{8}-[0-9a-f]{4}-[1-8][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000|ffffffff-ffff-ffff-ffff-ffffffffffff)$/i;
```
mjewildnhs: We are validating this in the sandbox but [my comment](https://github.com/NHSDigital/nhs-notify-app-response/pull/236#discussion_r3844884872) on the app-response PR meant to add it to the production code. 
Like there is for create message: https://github.com/NHSDigital/communications-manager-api/blob/master/proxies/shared/resources/jsc/Messages.Create.Validate.js#L38
I don't think we need any validation in the sandbox.

specification/schemas/responses/RecipientResponse.yaml
mjewildnhs: redocly is kicking out a warning since this file was renamed.
I think we'll need to rename the file for `repos/communications-manager-api/specification/schemas/components/RecipientResponse.yaml` to `RecipientResponseCallback` or something, preserving the title.
```
> communications-manager@0.0.1 publish
> mkdir -p build && redocly bundle specification/communications-manager.yaml --dereferenced --remove-unused-components --ext json | poetry run python scripts/set_version.py > build/communications-manager.json

bundling specification/communications-manager.yaml...
[1] specification/callbacks/recipient_response.yaml:32:15 at #/requestBody/content/application~1vnd.api+json/schema/properties/data/items

Two schemas are referenced with the same name but different content. Renamed RecipientResponse to RecipientResponse-2.

30 |             minItems: 1
31 |             items:
32 |               $ref: ../schemas/components/RecipientResponse.yaml
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
33 | responses:
34 |     '202':

Warning was generated by the bundler rule.
```

mjewildnhs:
Happy with the changes.
Leaving approval off as I suspect further changes to get the build/tests working when app-response routing resolved.
One outstanding [comment](https://github.com/NHSDigital/communications-manager-api/pull/1090#discussion_r3843134617) is around tests which we'll have a bit more clarity on when we get things working.

tests/api/message_responses/test_200_success.py
rhyscoxnhs: The real dev environment is also tested, we seed the data needed for the test ahead of a test run now

sandbox/app.js
```diff
@@ -22,12 +22,13 @@ function setup(options = {}) {
...
 function start(options = {}) {
     const server = app.listen(options.PORT || 9000, () => {
+        const address = server.address();
         log.info(JSON.stringify({
             timestamp: Date.now(),
             level: "info",
             app: app.locals.app_name,
             msg: "startup",
-            server_port: server.address().port,
+            server_port: address ? address.port : null,
```
rhyscoxnhs: This fixes some flakes with the unit tests, without this they sporadically fail in CI

tests/api/message_responses/test_404.py
```diff
@@ -0,0 +1,30 @@
...
+from lib.fixtures import *  # NOSONAR
+
+# ref has no app-response backend; the endpoint is deliberately disabled there
+pytestmark = pytest.mark.skipif(
+    os.environ.get("API_ENVIRONMENT") == "ref",
+    reason="message-responses endpoint is not available in ref"
+)
+
+
+@pytest.mark.devtestonly
```
Ian-Hodges: Seems a shame to only run these new tests against internal-dev. Any reason they can't?

Ian-Hodges: Just noticed Mike's comment above.

rhyscoxnhs: In order for these tests to work we need to seed the environment with long lived test data. My feeling is that dev is the only environment where this is suitable.

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
```diff
@@ -0,0 +1,31 @@
...
+        <Name>target.copy.pathsuffix</Name>
+        <Value>false</Value>
+    </AssignVariable>
+    <AssignVariable>
+        <Name>requestpath</Name>
+        {% if ENVIRONMENT_TYPE == 'sandbox' %}
+        <Template>/api/v1/message-responses/{data.messageId}</Template>
+        {% else %}
+        <Template>/api/response/{data.messageId}</Template>
+        {% endif %}
```
Ian-Hodges: I think it would be nice for the sandbox to be as similar to the real backend as possible. Is there anything stopping us from changing sandbox so it responds to the same path?

specification/responses/5xx/message_responses/500_MessageResponsesInternalServerError.yaml
```diff
@@ -0,0 +1,20 @@
+description: |+
```
Ian-Hodges: This isn't being used?

specification/responses/5xx/message_responses/500_MessageResponsesInternalServerError.yaml
Ian-Hodges: This isn't being used anywhere. Was it meant for specification/endpoints/get_responses.yaml?

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
rhyscoxnhs: Fixed

specification/responses/5xx/message_responses/500_MessageResponsesInternalServerError.yaml
rhyscoxnhs: Fixed

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
```diff
@@ -18,6 +18,9 @@
     <Set>
         <Headers>
             <Header name="X-Correlation-Id">{backendCorrelationId}</Header>
+            {% if ENVIRONMENT_TYPE != 'sandbox' %}
```
mjewildnhs: Why is this not included for sandbox tests?
I can see it was added under the "Test fixes" commit - so I'm assuming it breaks something but I don't see why it should.

Ian-Hodges (approved):

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
rhyscoxnhs: It won't error in sandbox (the policy has IgnoreUnresolvedVariables=true).

The guard is there because `client_id` is only populated by OauthV2.VerifyAccessToken, which we don't run in sandbox (see the `{% else %}` branch in `Partial.Target.PreFlowRequest.xml` — sandbox uses the mock Prefer-header fault steps instead of OAuth).

In sandbox there's no client id to send, and nothing consumes the header anyway (the mock has no authoriser — it only matters to the real app-response backend). 

It also keeps this in line with `AssignMessage.AuthenticationDetails`, which is already sandbox-guarded on this flow and every other endpoint. Without the guard we'd just forward an empty NHSD-Client-Id header into the mock.

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
```diff
@@ -0,0 +1,27 @@
...
+    </AssignVariable>
+    <AssignVariable>
+        <Name>requestpath</Name>
+        <Template>/api/response/{data.messageId}</Template>
+    </AssignVariable>
+    <Set>
+        <Headers>
+            <Header name="X-Correlation-Id">{backendCorrelationId}</Header>
+            {% if ENVIRONMENT_TYPE != 'sandbox' %}
+            <Header name="NHSD-Client-Id">{client_id}</Header>
```
cgitim: We use `X-APIM-Application-Id` to check client authentication.

tests/api/message_responses/test_400.py
```diff
@@ -0,0 +1,35 @@
+import os
```
cgitim: I think we need a 403 test as well?

proxies/shared/resources/jsc/helpers/validationChecks.js
```diff
@@ -57,6 +57,17 @@ const validateUuid = (errors, fieldValue, fieldPointer) => {
     return true
 }
 
+const validateUuidParameter = (errors, fieldValue, parameter) => {
```
cgitim: If the error is "The messageId path parameter is not a valid UUID." then the function should not be generically called `validateUuidParameter()`. Maybe `validateUuidMessageId()`?

cgitim (dismissed):

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
rhyscoxnhs: Fixed

proxies/shared/policies/AssignMessage.MessageResponses.Get.Request.xml
rhyscoxnhs: Fixed

tests/api/message_responses/test_400.py
rhyscoxnhs: Fixed

proxies/shared/resources/jsc/helpers/validationChecks.js
rhyscoxnhs: Fixed

specification/responses/5xx/message_responses/500_MessageResponsesInternalServerError.yaml
```diff
@@ -0,0 +1,20 @@
...
+  An error has occurred that is stopping your request from being processed.
+
+  The following errors can occur:
+
+  | Error code                 | Title                    | Description                                                 |
+  | -------------------------- | ------------------------ | ----------------------------------------------------------- |
+  | `CM_TOO_MANY_RESPONSES`    | Too many responses       | There are too many responses to return.                     |
+  | `CM_INTERNAL_SERVER_ERROR` | Error processing request | There was an internal error whilst processing this request. |
```
mjewildnhs: We took `CM_INTERNAL_SERVER_ERROR` out and need to remove it from the docs.
I think we can probably use the standard 500 template for this now?

specification/schemas/enums/ErrorMessageResponsesInternalServerError.yaml
```diff
@@ -0,0 +1,6 @@
+title: Enum_Error_MessageResponsesInternalServerError
+type: string
+enum:
+  - CM_TOO_MANY_RESPONSES
+  - CM_INTERNAL_SERVER_ERROR
```
mjewildnhs: Again `CM_INTERNAL_SERVER_ERROR` can come out

mjewildnhs (approved):
1 more minor comment to address - approving as I don't need to re-review

specification/responses/5xx/message_responses/500_MessageResponsesInternalServerError.yaml
rhyscoxnhs: Fixed

specification/schemas/enums/ErrorMessageResponsesInternalServerError.yaml
rhyscoxnhs: Fixed

